Issue #4, opened 2023-12-12
Implement the `Text Quote Selector` model that will match the W3C spec


oleksandr-danylchenko:
## Current State
During the walk through the lib code, I found the new [`TextSelector`](https://github.com/recogito/text-annotator-js/blob/8ccace6254f599d25ebb6ddbac8c69fa571c1cf0/packages/text-annotator/src/model/TextAnnotation.ts#L9-L13) added during the [internal data model shift](https://matrix.to/#/!PKeiixCnYWBHzKrJUT:gitter.im/$f2tJSZNtj3e58bSMZKGs1L5GqKLkjo4_O1682nz2KRY?via=gitter.im&via=messenger.ukp.informatik.tu-darmstadt.de)
It covers the functionality of both the [`Text Quote Selector`](https://www.w3.org/TR/annotation-model/#text-quote-selector) and the [`Text Position Selector`](https://www.w3.org/TR/annotation-model/#text-position-selector) in a more concise format of a single object:
https://github.com/recogito/text-annotator-js/blob/8ccace6254f599d25ebb6ddbac8c69fa571c1cf0/packages/text-annotator/src/model/TextAnnotation.ts#L15-L27

## Issue Description
Unfortunately, its _`Text Quote Selector` representation_ is only partial now, utilizing the [`exact`](https://www.w3.org/TR/annotation-model/#text-quote-selector:~:text=associated%20with%20it.-,exact,-Property) prop, _without capturing the [`prefix`](https://www.w3.org/TR/annotation-model/#text-quote-selector:~:text=1%20exact%20property.-,prefix,-Property) and [`suffix`](https://www.w3.org/TR/annotation-model/#text-quote-selector:~:text=more%20than%201.-,suffix,-Property)_. That could be sufficient if we know that the `target`'s `source` is cemented and will never change. But in our use case, we expect that the _`source` is mutable_ and we'll need to _"migrate"/"restore" the existing annotations on the updated content_. _Relying on the `exact` match can be deceiving during that process._ 

### Broken behavior example
1. A user selects the word "cat" in the 1st paragraph, within the sentence "walking cat in the park"
2. Another paragraph is added at the start with the sentence "my cat really likes playing with toys"
3. During the "migration" process we'll look for the `exact` match in the updated `source` and will add the annotation to the new sentence with the word "cat" in it. The original annotation will get lost

## Why Change is Needed
[...]

PR #5, opened 2023-12-12
Implemented capturing of the `prefix` & `suffix` for the selector to match the `Text Quote Selector`

oleksandr-danylchenko:
## Issue
This PR resolves the https://github.com/recogito/text-annotator-js/issues/4 issue

## Changes Made
In this PR I implemented capturing of the `prefix` & `suffix` properties to match the `Text Quote Selector` spec. For both of them I use the `Range` class that captures ±10 characters on the `start`/`endContainer` respectively. Then the textual content is simply extracted from the using the `.toString()`

## Refinement that can be made on demand
Another way to refine the `prefix`/`suffix` selection is to go to the `prev`/`nextSibling` when there are not enough characters to capture on `start`/`endContainer`. I started to work in that direction first but realized that it adds a serious complexity overhead, solving the edge case that hasn't blocked anyone yet 🤷🏻‍♂️ . So decided to go with a simpler approach for now 

## Demo

https://github.com/recogito/text-annotator-js/assets/68850090/0d0e3f77-0911-47a5-a95b-90033256c3c8

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -1,39 +1,42 @@
 import { Origin, type User } from '@annotorious/core';
 import { v4 as uuidv4 } from 'uuid';
 import type { TextAnnotatorState } from './state';
-import type { TextSelector, TextAnnotationTarget } from './model';
+import type { TextSelector, TextAnnotationTarget, TextSelectorQuote } from './model';
 import { trimRange } from './utils';
 
+const rangeToQuote = (range: Range): TextSelectorQuote => {
+  /**
+   * Captures the prefix and suffix snippets of the selection to match the `Text Quote Selector` spec
+   * @see https://www.w3.org/TR/annotation-model/#text-quote-selector
+   */
+  const snippetLength = 10;
+
+  const rangePrefix = document.createRange();
+  rangePrefix.setStart(range.startContainer, Math.max(range.startOffset - snippetLength, 0));
+  rangePrefix.setEnd(range.startContainer, range.startOffset);
+
+  const rangeSuffix = document.createRange();
+  rangeSuffix.setStart(range.endContainer, range.endOffset);
+  rangeSuffix.setEnd(range.endContainer, Math.min(range.endOffset + snippetLength, range.endContainer.textContent.length));
+
+  return {
+    exact: range.toString(),
+    prefix: rangePrefix.toString(),
+    suffix: rangeSuffix.toString()
+  }
+}
+
 export const rangeToSelector = (range: Range, container: HTMLElement, offsetReferenceSelector?: string): TextSelector => {
-  const offsetReference: HTMLElement = offsetReferenceSelector ? 
-    (range.startContainer.parentElement as HTMLElement).closest(offsetReferenceSelector) : container;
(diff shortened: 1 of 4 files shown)


rsimon:
That looks great, many thanks! I'll review in a bit more depth later. But looks as if all is fine mostly 👍 

Just one mini remark: I saw you introduced a `TextSelector` type. While I'm not opposed to that, I think moving `quote` 'downwards' one level might currently break downstream code in the TEI and PDF annotator implementations. Perhaps we can keep all properties (exact, prefix, suffix) at the same level for now, just to make sure other stuff doesn't break.

oleksandr-danylchenko:
> . Perhaps we can keep all properties (exact, prefix, suffix) at the same level for now, just to make sure other stuff doesn't break.

No problem 👌🏻

rsimon (approved):
This one looks good - many thanks!

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -1,30 +1,50 @@
...
+      ? (range.startContainer.parentElement as HTMLElement).closest(offsetReferenceSelector)
+      : container;
+
+  // Helper range from the start of the contentNode to the start of the selection
+  const rangeBefore = document.createRange();
   rangeBefore.setStart(offsetReference, 0);
   rangeBefore.setEnd(range.startContainer, range.startOffset);
 
-  const quote = range.toString();
+  const quoteSelector = rangeToQuoteSelector(range);
```
rsimon, line 37: BTW determining prefix/suffix could actually be simplified - the `rangeBefore` already contains all text before the quote. A `rangeAfter` could be constructed in a similar way.

packages/text-annotator/src/model/TextAnnotation.ts
```diff
@@ -12,10 +12,26 @@ export interface TextAnnotationTarget extends AnnotationTarget {
 
 }
 
-export interface TextSelector {
+/**
+ * Matches the `Text Quote Selector` spec
+ * @see https://www.w3.org/TR/annotation-model/#text-quote-selector
+ */
+export interface TextQuoteSelector {
```
rsimon, line 19: A bit like my comment concerning the LifecycleInfo type: I don't think there's a benefit to split the TextSelector into two parts. They won't ever be handles in separate in the core model. Whether WebAnno treats them differently is another matter, and IMO not necessarily a concern for the core model.

rsimon (approved):

rsimon:
Just a note to say that I made a few changes.
- The prefix/suffix computation is now based on ranges that span the complete container
- I moved the computation into a separate helper function
- __The function is currently no longer used in the core.__ On second thought, I believe that generating prefix/suffix should live in the W3CAdapter after all. Since the core model Selector will always include the Range object, it makes no difference if computation is done in the core or the adapter. IMO it feels cleaner this way, and it doesn't impact the datastructure that's already used in other apps I've been building.

I'll look into the adapter code next. Hopefully later today.

oleksandr-danylchenko:
> Since the core model Selector will always include the Range object, it makes no difference if computation is done in the core or the adapter. 

Yep, that makes sense to me now. We don't need the prefix/suffix to show the highlights in the Core, the `range` will do that all 👌🏻

PR #8, opened 2023-12-17
Fixes/revisions to original PR#5

rsimon:
## In this PR

This PR makes some revisions to the original PR #5, which aims to add support for `prefix` and `suffix` values in the text selector.

Note that this PR __keeps prefix and suffix out of the core model.__ They will be tied in only at the level of the W3CFormatAdapter, which is coming up in another PR.

### DIFF
--- a/packages/text-annotator/src/utils/getContext.ts
+++ b/packages/text-annotator/src/utils/getContext.ts
@@ -1,0 +1,31 @@
+export const getContext = (
+  range: Range, 
+  container: HTMLElement, 
+  length = 10,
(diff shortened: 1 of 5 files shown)


packages/text-annotator/src/utils/getContext.ts
oleksandr-danylchenko, line 1: Very nice! Plus, the "wide" `rangeBefore`/`rangeAfter` will be able to capture the siblings' content as well. I didn't think about such an approach 🤔

PR #9, opened 2024-01-04
Applied unified indentation/punctuation formatting

oleksandr-danylchenko:
## Issue
This PR is meant to apply formatting fixes suggested by IDE to prevent diff churn in the following PRs. When I was working on the https://github.com/recogito/text-annotator-js/pull/7, I faced a ton of style-related changes that were just needlessly bloating the "Files changed".

## Changes Made
I applied IDE's autoformatting on the files I'm about to touch in the following PR where I'm about to add the W3C adapter. It mostly removed redundant spaces and unified usage of the `;`.

Issue #6, opened 2023-12-13
Implement default W3C format adapter


oleksandr-danylchenko:
## Current State
According to the [discussion in Gitter](https://matrix.to/#/!PKeiixCnYWBHzKrJUT:gitter.im/$speczEklGPPVks1JeknwjAupkUvjjsFBiZxPVoLziZw?via=gitter.im&via=messenger.ukp.informatik.tu-darmstadt.de):
> Annotorious v3 internally uses a model that's aligned with W3C, but it's not exactly identical. Consider it a slightly derived model that is meant for representing internal state: it's designed to work fast, with no need for constant runtime crosswalks.
> 
> In addition, there are "adapters", that will handle the crosswalking to W3C (or potentially other formats - COCO etc.).
> 
> So far, Annotorious v3 ships with a default [W3C format adapter implementation](https://github.com/annotorious/annotorious/blob/3.x/packages/annotorious/src/model/w3c/W3CImageFormatAdapter.ts). The adapter handles the crosswalking for everyone who wants to use the WebAnno spec. But it crosswalks only "at the edges" of the library, as it were.

## Issue Description
> **The crosswalk adapter is not yet implemented in the Text Annotator! Only in Annotorious.**

## Why Change is Needed
Such an adapter for the `TextAnnotator` would be useful for the users that need to persist the annotations in the classic W3C WebAnnotaion Model format and utilize the [annotation-protocol](https://www.w3.org/TR/annotation-protocol/).

## Possible Solutions
Create a built-in adapter called `W3CTextFormatAdapter` with `parseW3CTextAnnotation` & `serializeW3CTextAnnotation`. Such an adapter can be opted-in by a library user if they want to follow the W3C spec on their side.
The adapter should support mapping for such properties as: 
- `@context`
- `id`
- `type`
- `body` and its sub-properties*
- `target` and its sub-properties**

But all the other possible properties from the W3C WebAnno spec like: `generator`, `generated`, `audience`, `stylesheet` should be just recorder/serialized w/o any changes. Because they are not being used during the lifecycle of the package/

_*`body.source`, `body.rights` will be removed from the W3C annotation payload on parsing because there is no match for them in `AnnotationBody`
**`target.styleClass`,`target.accessibility`, `target.state` will be removed from the W3C annotation payload on parsing because there is no match for them in `TextAnnotationTarget`_

## Additional Information
[...]

PR #10, opened 2024-01-04
Implemented the default W3C format adapter for `TextAnnotation`

oleksandr-danylchenko:
## Issue
This PR resolves the https://github.com/recogito/text-annotator-js/issues/6 issue

## Changes made
In this PR I implemented the default W3C format adapter `W3CTextFormat` which is a close relative of the `W3CImageFormat`. The change can be described in a few major steps:
1. Added [`W3CTextAnnotation`](https://github.com/oleksandr-danylchenko/text-annotator-js/blob/w3c-text-annotations-adapter/packages/text-annotator/src/model/w3c/W3CTextAnnotation.ts) type definition for the expected `W3C` format of the text notes. They explicitly support `TextQuoteSelector` & `TextPositionSelector`, as those are the only selectors we use in the package.
2. Made the `reviveTargetRange` function which obtains the `Range` from the `container` based on the `start`/`end` props. That function was mostly extracted from the broader `reviveTarget`, which creates an entire `target` object.
3. Added [`W3CTextFormatAdapter`](https://github.com/oleksandr-danylchenko/text-annotator-js/blob/w3c-text-annotations-adapter/packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts), which is responsible for all the transforms to/from the W3C format of the text annotations.
3.1. on parse the `selector` is gathered from the combination of the `TextQuoteSelector` & `TextPositionSelector`. If either quote or position is missing - the error will be returned

## Related PRs
**- This PR is based and depends on previously made https://github.com/recogito/text-annotator-js/pull/9 where I resolved formatting issues**
- This PR depends on the Annotorious PR https://github.com/annotorious/annotorious/pull/347 where I made `W3CSelector` abstract which made it possible to declare the `W3CTextAnnotationTarget` with `selector: W3CTextSelector | W3CTextSelector[]` props. Otherwise, it would complain about not having a required `value` prop: 
 

## Following changes
In the following PR, I plan to add tests for the for the adapter using `vitest` + `jsdom`


rsimon:
Thanks, I'll try to take a look later today tomorrow. For now: please don't change the coding style of the project. I'm happy to discuss changes further down the line, perhaps in line with a prettier rule or similar. But for now, it makes the PRs much harder to read.

oleksandr-danylchenko:
> Thanks, I'll try to take a look later today tomorrow. For now: please don't change the coding style of the project. I'm happy to discuss changes further down the line, perhaps in line with a prettier rule or similar. But for now, it makes the PRs much harder to read.

I specifically extracted allll those annoying changes into a separate PR - https://github.com/recogito/text-annotator-js/pull/9
After it'll get merged, this PR will contain only business-logic changes

oleksandr-danylchenko:
It's such a bummer that it's impossible to use the "stacked PRs" approach between a fork and the main repo. Otherwise, diffs would be much cleaner

PR #14, opened 2024-01-19
Added `types` definition into `exports`

oleksandr-danylchenko:
## Issue
When I started using the `@recogito/react-text-annotator` dependency in the project with the `Yarn v.1.22.19` & `Node v.20` I was hella surprised that none of the types from the `@recogito/text-annotator` were processed by the Typescript:
image
Also, no proper hits were provided for the `TextAnnotator` component: 
image
Although IDE could navigate through the definition files they definitively were present in the `node_modules` 🤔 

## Changes made
The issue I found was that the `@recogito/text-annotator` didn't have the `types` property specified within the `exports` like in the `@recogito/react-text-annotator`: https://github.com/recogito/text-annotator-js/blob/1df4dd1123d637aa244070dea01ea1023b863951/packages/text-annotator-react/package.json#L17-L21
After I added it - everything started to work as intended, including automatic code suggestions ✅ 
image

I suppose that with such a Yarn + Node combination, the `exports` prop takes precedence over the top-level `main` & `types`...

### DIFF
--- a/packages/extension-tei/package.json
+++ b/packages/extension-tei/package.json
@@ -19,6 +19,7 @@
   ],
   "exports": {
     ".": {
+      "types": "./dist/src/index.d.ts",
       "import": "./dist/text-annotator-tei.es.js",
       "require": "./dist/text-annotator-tei.umd.js"
     },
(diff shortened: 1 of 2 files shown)


rsimon:
Yikes - you are right of course! I totally missed it. Thanks for the fix!

PR #17, opened 2024-01-22
Added style props types for the `TextAnnotation`

oleksandr-danylchenko:
In our use-case, we would like to choose different colors for the highlights made by a user. And the [`styleClass` fits](https://www.w3.org/TR/annotation-model/#styles) it perfectly. We can apply a class to the annotation on creation. And  then process it in the `TextAnnotator` `style` callback (roughly) like this:
https://github.com/recogito/text-annotator-js/blob/8bb8af986bd90e730c22f13bcd20e83060573b7e/packages/text-annotator/src/TextAnnotatorOptions.ts#L15
```ts
const annotationsStyle = (annotation: TextAnnotation): DrawingStyle => {
	if (annotation.target.styleClass === 'highlight-1') {
		return {
			fill: theme.annotations....
			fillOpacity: 0.2
		};
	} else {
	}
};
```

### DIFF
--- a/packages/text-annotator/src/model/core/TextAnnotation.ts
+++ b/packages/text-annotator/src/model/core/TextAnnotation.ts
@@ -10,6 +10,8 @@ export interface TextAnnotationTarget extends AnnotationTarget {
 
   selector: TextSelector;
 
+  styleClass?: string;
+
 }
 
 export interface TextSelector {
--- a/packages/text-annotator/src/model/w3c/W3CTextAnnotation.ts
+++ b/packages/text-annotator/src/model/w3c/W3CTextAnnotation.ts
@@ -4,11 +4,15 @@ export interface W3CTextAnnotation extends W3CAnnotation {
 
   target: W3CTextAnnotationTarget | W3CTextAnnotationTarget[];
 
+  stylesheet?: W3CAnnotationStylesheet;
+
 }
 
 export interface W3CTextAnnotationTarget extends W3CAnnotationTarget {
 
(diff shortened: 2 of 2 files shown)


oleksandr-danylchenko:
I intentionally didn't add those props to the Annotorious package, as the `styleClass` won't be used there (at least there are no such use-cases for now)

rsimon:
IMO it's ok to add this to the W3C annotation. But not to the core. In the core, styling isn't considered to be a special, hard-wired property of the annotation. (Styling could be based on all sorts of state variables.) The styling mechanism in Annotorious is already implemented via the `style` function, and should be kept the same for the TextAnnotator.

oleksandr-danylchenko:
> IMO it's ok to add this to the W3C annotation

Okay, then it might be worth specifying it in Annotorious for the sake of a "single source of truth". 

> In the core, styling isn't considered to be a special, hard-wired property of the annotation

I intended to specify just the `styleClass` prop to help IDE's intelisense, but it _won't be processed in the lib directly_:
image
Do you think it might feel misleading for the lib's users? Should I just extend the `TextAnnotation` locally within a specific app?

rsimon:
Yes, I think it should stay out of the core. In the longer run (hopefully not too far future...) both the `selectAction` and the `style` function would be affected by the adapter. Which means you would get a `W3CTextAnnotation` as an argument to your style (or pointerAction) function. This way, intellisense will work again, and we could limit `styleClass` the just the W3CTextAnnotation.

rsimon:
Thanks for the discussion & the useful addition!

PR #19, opened 2024-01-24
Unified `style` callback type

oleksandr-danylchenko:
## Issue
When I wanted to add styling for individual annotations based on their `styleClass` and selection state - I realized that the 2nd `selected` argument is not specified in the type def for the `style` prop 👀 
Although it's being passed into the `drawingStyle` function: https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/highlight/HighlightPainter.ts#L32

## Changes made
In this PR I tried to resolve a somewhat inconsistent definition of the `style` with defining a single `HighlightPainterStyle` type


rsimon:
Sorry, this is some more work in progress. The second argument will change. It's currently just a boolean flag (isSelected). But will turn into an object which includes the following:

- selection (boolean)
- mouse hover (boolean)
- custom (an object that can be manipulated by the application - this is to support use cases where, e.g., an application has a list of annotations outside the text/image, and wants to style annotations according to whether a user selects them in the list.)

oleksandr-danylchenko:
Won't having a single shared type help in that transition? I suppose it'll be easier to replace the 2nd argument type just in the `HighlightPainterStyle` instead of a bunch of other places 🤔

packages/text-annotator/src/highlight/HighlightPainter.ts
```diff
@@ -17,11 +17,13 @@ export type HighlightPainter = {
     bg: CanvasRenderingContext2D,
     fg: CanvasRenderingContext2D,
     isSelected?: boolean,
-    style?: DrawingStyle | ((annotation: TextAnnotation, selected?: boolean) => DrawingStyle)
```
oleksandr-danylchenko, line 20: `DrawingStyle` comes from the `A9S`, but for the text annos, there's ` HighlightStyle` type defined below: https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/highlight/HighlightPainter.ts#L31
https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/highlight/HighlightPainter.ts#L44-L56

rsimon:
I'm not sure I understand what you're trying to do. (But didn't have time to read in detail.) Is this just about adding the underline style option, to extend the default DrawingStyle? If this is to shorten the combination of (annotation => DrawingStyle) | DrawingStyle type: I think it's actually an advantage to explicitly spell this out. A style can be be a fixed object, or a function that generates such an object on demand. Wouldn't want to mask it with an extra type TBH.

oleksandr-danylchenko:
>  If this is to shorten the combination of (annotation => DrawingStyle) | DrawingStyle type: I think it's actually an advantage to explicitly spell this out.

Something like this. I noticed that `style` type was inconsistently specified in multiple places. Some of that looked like:
```ts
DrawingStyle | ((annotation: TextAnnotation) => DrawingStyle)
``` 
but others like:
```ts
DrawingStyle | ((annotation: TextAnnotation, selected?: boolean) => DrawingStyle)
```

Although, the callback definition _was always used as: `(annotation: TextAnnotation, selected?: boolean) => DrawingStyle`_: https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/highlight/HighlightPainter.ts#L32

So by adding a single `HighlightPainterStyle` I tried to _resolve that inconsistency_. Plus the type definition for the `TextAnnotator` component _was incomplete_: `DrawingStyle | ((annotation: TextAnnotation) => DrawingStyle)`, which didn't allow me to style annotations based on the selection state:
https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator-react/src/TextAnnotator.tsx#L14

oleksandr-danylchenko:
> Is this just about adding the underline style option, to extend the default DrawingStyle?

No, the `HighlightStyle` was already used to definite the `style` type for the text annotations: 
https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/highlight/HighlightPainter.ts#L31
This makes the `style?: DrawingStyle | ((annotation: TextAnnotation) => DrawingStyle); ` _type incorrect_, as it doesn't return the `HighlightStyle`, but the A9S's `DrawingStyle`

oleksandr-danylchenko:
Oh... Is the `underline` rendering supported now?... It doesn't seem to be applied to the anno 🤔 
image
image

rsimon:
Yes, I think that might not be implemented yet :-)

rsimon:
Hi, just picking up on this...

I'm about to add a second rendering implementation, based on the [CSS Custom Highlight API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API). (It will simplify the code, and resolve a nasty rendering issue you may have encountered: annotations drift out-of-sync with the text during scrolling. It happens mostly for large text documents and is especially visible on mobile browsers.) Since Custom Highlights do not support underlines, I'd vote for removing the HighlightStyle, and sticking with the normal `DrawingStyle` from Annotorious. Does this make sense?

oleksandr-danylchenko:
Hello 🙌🏻 
Thanks for checking in

---

> I'm about to add a second rendering implementation, based on the [CSS Custom Highlight API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API).

That sounds a like great step toward the progressive library improvement. It's always nicer to use a native API instead of crafting custom workarounds 👍🏻 
Will the Highlight API rendering be an opt-in solution? Or does it mean the Canvas-based implementation will get deprecated and it won't receive fixed/updates then?
Unfortunately, the latter option doesn't sound so prominent for our project which needs to support some older browsers. And the Safari is the most severe pain point all the time:
image

---

> Since Custom Highlights do not support underlines ...

I'm not sure about that statement 🤔 
Because in the `::highlight()` styling doc they are using the `text-decoration: underline` property that's properly rendered in Chrome & Safari: 
image

rsimon:
Yes, the Custom Highlight renderer will be completely opt-in. Actually, my plan is to completely disable it by default. You'd switch it on with a new init option (something like `experimentalHighlightAPI: true`). And even then, the library itself would check if the device supports the API, and either use it or fall back to Canvas rendering.

Re underline: yes, `text-decoration: underline` works. But what I would have wanted was a bottom border underneath the highlight itself (as used to be [the default style in RecogitoJS](https://raw.githubusercontent.com/recogito/recogito-js/main/screenshot.png)). Alas, that's now out of the question.

Re browser support: I found that the Highlight API works well in Chrome and Safari, desktop/Android/iOS. Although on iOS I had to update to the latest OS release first. (I believe 17.2 was only released in December.) On our end, Firefox remains the key legacy scenarios, really.

oleksandr-danylchenko:
> Yes, the Custom Highlight renderer will be completely opt-in. Actually, my plan is to completely disable it by default. You'd switch it on with a new init option (something like experimentalHighlightAPI: true)

Great! It makes sense to me!

> Re underline: yes, text-decoration: underline works. But what I would have wanted was a bottom border underneath the highlight itself (as used to be [the default style in RecogitoJS](https://raw.githubusercontent.com/recogito/recogito-js/main/screenshot.png)).

Maybe the border style can be emulated by using the `text-underline-offset` and `text-decoration-color`. Playground - https://developer.mozilla.org/en-US/play?id=nYCSiAB9X5gJuEpKkB9wy6LlQp51l2UGHyZcVNOPkXDHQaOjg26ZScZrwTF8%2BIUvoZhNvoUyYVg%2Fz2BX

```css
::highlight(rainbow-color-1) {
  background-color: rgba(235, 149, 50, 0.3);
  text-decoration: underline;
  text-underline-offset: 6px;
  text-decoration-color: orange;
}
```
image

rsimon:
Nice! Yes, that might work :-)

oleksandr-danylchenko:
> Since Custom Highlights do not support underlines, I'd vote for removing the HighlightStyle, and sticking with the normal DrawingStyle from Annotorious.

Then we might need to keep the `HighlightStyle` around  😅

rsimon:
How about naming it TextDrawingStyle? And making it something like:

```ts
type TextDrawingStyle = DrawingStyle & {
  underline: string;
}
```

oleksandr-danylchenko:
I like the previous name `HighlightStyle` a bit more 🤔  
Because what we draw for the text is the _highlight_. And to precise with the props that type might look like:
```ts
export interface HighlightStyle extends Pick<DrawingStyle, 'fill' | 'fillOpacity'> {

  underlineStyle?: string;

  underlineColor?: number;

  underlineOffset?: number;

  underlineThickness?: number;

}
```
- Inherits only the `fill` & `fillOpacity`. Because set of the `DrawingStyle` props  are stroke-related and won't be applied to the text
- I think we should expose more properties to allow more fine-tuned styling for the underline

rsimon:
Yep, looks good. I believe I'd prefer all the props to sit at the same level TBH.

oleksandr-danylchenko:
Closed in favor of creating clearer Issue + PR combination using the newer `HighlightStyle` type

PR #20, opened 2024-01-24
Added event listeners cleanup in the React example

oleksandr-danylchenko:
## Issue
I noticed that in the React example event listeners are only being added to the `anno`, but never cleaner. So it may lead to the memory leaks on `MockStorage` re-renders.

## Changes made
Added cleanup function for the existing event listeners. Hopefully, it'll help other devs to immediately use the same pattern in their projects too

### DIFF
--- a/packages/text-annotator-react/test/App.tsx
+++ b/packages/text-annotator-react/test/App.tsx
@@ -35,22 +35,25 @@ const MockStorage = () => {
   const anno = useAnnotator<RecogitoTextAnnotator>();
 
   useEffect(() => {
-    if (anno) {
-      anno.on('createAnnotation', (annotation: TextAnnotation) => {
-        console.log('create', annotation);
-      });
+    if (!anno) return;
 
-      anno.on('deleteAnnotation', (annotation: TextAnnotation) => {
-        console.log('delete', annotation);
-      });
-    
-      anno.on('selectionChanged', (annotations: TextAnnotation[]) => {
-        console.log('selection changed', annotations);
-      });
-    
-      anno.on('updateAnnotation', (annotation: TextAnnotation, previous: TextAnnotation) => {
-        console.log('update', annotation, previous);
-      });
+    const handleCreateAnnotation = (annotation: TextAnnotation) => console.log('create', annotation);
+    anno.on('createAnnotation', handleCreateAnnotation);
+
+    const handleDeleteAnnotation = (annotation: TextAnnotation) => console.log('delete', annotation);
+    anno.on('deleteAnnotation', handleDeleteAnnotation);
+
+    const handleSelectionChanged = (annotations: TextAnnotation[]) => console.log('selection changed', annotations);
+    anno.on('selectionChanged', handleSelectionChanged);
+
+    const handleUpdateAnnotation = (annotation: TextAnnotation, previous: TextAnnotation) => console.log('update', annotation, previous);
+    anno.on('updateAnnotation', handleUpdateAnnotation);
+
+    return () => {
+      anno.off('createAnnotation', handleCreateAnnotation);
+      anno.off('deleteAnnotation', handleDeleteAnnotation);
+      anno.off('selectionChanged', handleSelectionChanged);
+      anno.off('updateAnnotation', handleUpdateAnnotation);
     }
   }, [anno]);
 


oleksandr-danylchenko:
### Oh-oh... The duplicated event listeners addition may be the issue not only for the example... 

In my app that has 2 re-renders in the parent component of the `TextAnnotator` the `selectstart` event is being processed 2 times:
https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/SelectionHandler.ts#L64
image
So now, 2 annotations always painted over each other

oleksandr-danylchenko:
I realized why it works that way in my app!
image

Which brings a questions, why the condition to call the `createTextAnnotator` is `if (setAnno)` and not the `if (setAnno && !anno)`?.. 🤔 
I suppose it'll fix the issue with redundant creation of the duplicate annotator instances and event listeners

rsimon:
Good point. Although when the context (which provides setAnno) gets re-mounted, so should the annotation layer I guess. setAnno shouldn't normally change, because its from a state hook in the conetxt. Therefore I believe a cleaner fix would be to destroy anno in a cleanup function that gets returned in the effect hook?

I believe I added an anno.destroy() method meanwhile. That may not yet have existed at the time I started the React binding.

oleksandr-danylchenko:
Yeah, destroying the might be the best solution on re-renders. But there's a nuance 😆 
https://github.com/recogito/text-annotator-js/blob/054ca425b2957c9184565595ff40bfbd5c26a835/packages/text-annotator/src/TextAnnotator.ts#L86-L88

oleksandr-danylchenko:
I think I can try to work on the `destroy` method tomorrow. As the creation of duplicated annotations blocks me from moving forward 🤷🏻‍♂️ 
As I won't be able to predictably discard a highlight or delete it

oleksandr-danylchenko:
Or, if you already have a vision of how the `destroy` method should look like for the `TextAnnotator`, I can proceed with just a `ref` hack:
image

rsimon:
You can take a look at what Annotorious does here: https://github.com/annotorious/annotorious/blob/main/packages/annotorious/src/Annotorious.ts

It's basically a matter of housekeeping (like destroying the undo stack, unregistering listeners) plus reverting the DOM manipulations that were made. (Mainly the inserted Canvas elements.) I think both the SelectionHandler and the Highlighter would need their own destroy methods, and then these would be called from the Annotator.

oleksandr-danylchenko:
Anyway... I suppose this PR should be reviewed individually. As the change was only for the example and it fixes the duplicated events logs there

rsimon:
Great! And agreed - this PR looks good, independently of the discussion.

rsimon:
Just a quick heads-up: I need to make a few changes to the text annotator & will add a destroy method to the `highlightLayer`!

rsimon:
Ok, I believe most should be there now. Proper unregistering of listeners + destroying the highlight layer Canvases (not tested though ;-)

oleksandr-danylchenko:
> Just a quick heads-up: I need to make a few changes to the text annotator & will add a destroy method to the `highlightLayer`!

Thanks for the heads-up 🙌🏻 
Luckily, I haven't started working on the `destroy` method yet. Yesterday I got sworn into non-anno related tasks

oleksandr-danylchenko:
Oh-oh... There might be some bug in the highlighter resing handler now. As the handler gets fired a bazillion time despite the page being fully stating and I'm not even touching anything 👁️ 
image

oleksandr-danylchenko:
All those resize calls were coming from the `MutationObserver`!
image

oleksandr-danylchenko:
And it main issue is the `attributes` property in the `observe` config: 
image
After I removed it - the constant handler calls stopper

rsimon:
Ah, I think I see what's going on. The mutation observer triggers a redraw (as it should). But that also resets the Canvas elements' width and height attributes. Yes, removing the attributes will work. But probably not the final solution. In my use case, I need to redraw the canvas as parent elements in the layout change their size due to style prop changes. So I assume the mutation observer has to listen to all events (incl. attribute changes) but then ignore changes to the drawing canvases. Will require a bit more work...

rsimon:
Quick update on this one: the infinite loop is fixed. (Turned out the solution was easier than thought. I didn't have to do a resize + canvas reset. A simple redraw was totally sufficient, and doesn't cause a DOM mutation which would then throw the observer into an infinite loop.)

oleksandr-danylchenko:
Sounds great! Thanks!

PR #22, opened 2024-01-26
Added `destroy` method for the `TextAnnotator`

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/pull/20#issuecomment-1908416217

## Changes made
In this PR I:
- Added `destroy` method for the `SelectionHandler`
- Added `destroy` method for the `TextAnnotator`
- Migrated React example to the `StrictMode`. It helped to validate that even on 2 `useEffect` invocations - only a single annotator instance stays active

### DIFF
--- a/packages/text-annotator-react/src/TextAnnotator.tsx
+++ b/packages/text-annotator-react/src/TextAnnotator.tsx
@@ -1,10 +1,9 @@
 import { ReactNode, useContext, useEffect, useRef } from 'react';
 import { AnnotoriousContext, DrawingStyle, Filter } from '@annotorious/react';
-import { createTextAnnotator } from '@recogito/text-annotator';
 import type { TextAnnotation, TextAnnotatorOptions } from '@recogito/text-annotator';
+import { createTextAnnotator } from '@recogito/text-annotator';
 
 import '@recogito/text-annotator/dist/text-annotator.css';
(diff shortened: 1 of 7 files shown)


oleksandr-danylchenko:
Now it works fine in my app and there are no overlapping of the highlights 🙌🏻 
image

packages/text-annotator-react/src/TextAnnotator.tsx
```diff
@@ -4,6 +4,7 @@ import { createTextAnnotator } from '@recogito/text-annotator';
 import type { TextAnnotation, TextAnnotatorOptions } from '@recogito/text-annotator';
 
 import '@recogito/text-annotator/dist/text-annotator.css';
+import { textAnnotation } from '@recogito/text-annotator/dist/test/model/w3c/fixtures';
```
rsimon, line 7: That might be a copy-and-paste error.

rsimon (approved):
Looks good. Nice job also about refactoring the debounce utility!

packages/text-annotator-react/src/TextAnnotator.tsx
oleksandr-danylchenko: Exactly! Thanks for highlighting it!

rsimon:
Wonderful - thanks!

PR #30, opened 2024-02-05
Fixed non-inclusive `range` revival

oleksandr-danylchenko:
## Issue
When the `selector` `end` position matches the end of the `text` node - the revived range ends on the next node's start
image

---

I noticed it while working on omitting the not annotable elements. There made the revive method iterator to skip them:
```ts
const iterator = document.createNodeIterator(offsetReference, NodeFilter.SHOW_TEXT, (node) =>
	node.parentElement?.closest(notAnnotableSelector)
		? NodeFilter.FILTER_SKIP
		: NodeFilter.FILTER_ACCEPT
);
```
So instead of ending on the not annotable element, the range was ending on the next annotable element. Leading to unexpected range positioning:
image

## Changes Made
All the hurdle was caused by the non-inclusive comparison clause for the `end` position. It doesn't allow the range to capture a single `text` node parent

### DIFF
--- a/packages/text-annotator/src/state/reviveTarget.ts
+++ b/packages/text-annotator/src/state/reviveTarget.ts
@@ -51,7 +51,7 @@ export const reviveTarget = (
   while (n !== null) {
     const len = n.textContent.length;
 
-    if (runningOffset + len > end) {
+    if (runningOffset + len >= end) {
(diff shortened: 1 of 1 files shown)


rsimon:
Excellent - thanks!

PR #29, opened 2024-02-05
Fixed rendering of the pre-defined annos for the `text-annotator` example page

oleksandr-danylchenko:
## Issue
When I was working with the `text-annotator` example page - I noticed that none of the pre-defined annotations were rendered on the page 👁️👁️👁️

## Changes made
1. Added usage of the `annotations.w3c.json` instead of the deleted `annotations.json` one
2. Added usage of the `W3CTextFormat` adapter to properly parse the new json
3. Fixed selectors positions according to the page content. I just re-selected the same quote and applied the positions reported in the `updateAnnotation` handler

Issue #24, opened 2024-01-30
Support for the omission of the not annotable elements within the text


oleksandr-danylchenko:
## Current State
The Core [`TextAnnotation`](https://github.com/recogito/text-annotator-js/blob/80c2f9235ef27b2e57ff84a23d1f100a7b377e83/packages/text-annotator/src/model/core/TextAnnotation.ts#L3-L7) model is based on the fact that an annotation can have just a single `target`. Which matches the A9S's [`Annotation`](https://github.com/annotorious/annotorious/blob/b220e987b74dec2b2c19db45574a29e11c93a115/packages/annotorious-core/src/model/Annotation.ts#L3-L17) where target is an image region (_singular_).
I understand that having a _single `target`_ with a _single `range`_ also makes sense in the context of most browsers returning just a [single range on selection](https://javascript.info/selection-range#selection). Such an approach works perfectly well as long as the annotable content is just text with images.

## Issue
Usually, the content of our pages may contain visual elements with text, but they should not be annotable: 
<img src="https://github.com/recogito/text-annotator-js/assets/68850090/49085a2b-61fd-4a75-b38a-541cdb111fce" width="333" />
Or a page may contain interactive elements that also shouldn't be annotable: 
<img src="https://github.com/recogito/text-annotator-js/assets/68850090/d9a6ec1a-6657-4bd7-b563-de32c50fd7d2" width="333" />
_But, unfortunately, the `SelectionHandler` doesn't support handling multiple ranges, either by creating multiple `target`s or even multiple annotations._ 😔

## Possible Solutions

1. Using multiple `target`s per single annotation. For me, it sounds like the most optimal one. It matches the `target`'s meaning from spec. And it allows working with an existing popup conveniently. You can click any of the anno's targets and you'll still be addressing the same parent instance. Which is useful for changing the highlighting color/canceling it or adding a note.
2. Creating multiple annotations with a single target. Each will address a range before/after the not annotable element. But for me, it doesn't sound that convenient. I see that in most cases we expect a single annotation to be selected at a time. Plus, managing a shared popup for different annos will be quite painful.

However, I see the main _bottleneck_ in both cases _is the performance_. As the annotation is created/updated on the `selectionchange` event, it would be quite burdensome for the browser to split a single selection range into multiple targets or create multiple annotations. 
The more appropriate place for that could be the `pointerup` event handler 🤔. But in that case, it's not clear whether we should remove that `target` generation from the selection event handler entirely 🤔

## Questions
[...]

PR #31, opened 2024-02-05
#365 Added support for multiple `targets` selection in the `text-annotator`

oleksandr-danylchenko:
Relations
This PR is fully based on the https://github.com/recogito/text-annotator-js/pull/28 and https://github.com/recogito/text-annotator-js/pull/29 PRs. It includes all of their changes as well. So I would recommend starting from them first.
Also, to support such a change the A9S PR https://github.com/annotorious/annotorious/pull/369 is required to be in place.

## Issue
This PR resolves the https://github.com/recogito/text-annotator-js/issues/24 issue and helps to achieve the omission of not annotable elements within the text.

## Demo
image

## Changes made
In this PR, following the https://github.com/annotorious/annotorious/pull/367, I replace the usage of a single `target` with multiple `targets`. Each of those targets matches the selection range on an annotable element!

A single selection range gets split when an element with the `data-not-annotable="true"` attribute is found within it. Also, the not annotable elements are not included in the position counting! It's convenient to skip elements with dynamic content.  

To support such a behavior the most significant changes were introduced to the `spacialTree`, `TextAnnotationState`, and `SelectionHandler`


packages/text-annotator/src/state/reviveTarget.ts
```diff
@@ -1,57 +1,55 @@
...
-  if (!offsetReference)
-    return target;
-
-  const iterator = document.createNodeIterator(container, NodeFilter.SHOW_TEXT);
+export const reviveTargetRange = (start: number, end: number, offsetReference: HTMLElement): Range => {
+  const iterator = document.createNodeIterator(offsetReference, NodeFilter.SHOW_TEXT, (node) =>
+    node.parentElement?.closest(notAnnotableSelector)
+      ? NodeFilter.FILTER_SKIP
+      : NodeFilter.FILTER_ACCEPT
+  );
```
oleksandr-danylchenko, line 19: Skips the not annotable elements (possibly dynamic) for a proper counting

packages/text-annotator/src/state/spatialTree.ts
```diff
@@ -19,35 +23,40 @@ export interface IndexedHighlightRect {
...
+  target: {
+
     id: string;
 
+    annotation: string;
+
     rects: DOMRect[];
 
-  }
+  };
```
oleksandr-danylchenko, line 34: Now the `target` is the main entity for the indexed highlight rects. Because each `target` has its own set of `rects` and each annotation can have _multiple targets_

oleksandr-danylchenko:
Closed in favor of the #34

PR #33, opened 2024-02-12
Removed the unused `getClientRectsPonyfill`

oleksandr-danylchenko:
## Changes Made

I noticed that since September the Firefox-specific handling for obtaining client rects has been deactivated. And if since then there were no issues w/ Firefox - I think that we can remove the unused ponyfill


rsimon:
I'll leave this open for now. Need to revisit. The problem is that Firefox deals with highlight computation differently (in a bad way) as soon as block-level elements are involved. Leading to the following difference:

### Chrome (= the behavior we want)
Bildschirmfoto 2024-02-18 um 09 34 21

### Firefox (bad)
Bildschirmfoto 2024-02-18 um 09 36 09

The ponyfill temporarily modifies the DOM, inserts SPANs, and then measures highlights from the SPANs rather than the Ranges. This is a pretty heavy computation, and occasionally introduces unwanted side effects on the Range object. Because of this, I currently disabled it. But it remains an open issue.

oleksandr-danylchenko:
That's crazy 🤯 
Thanks for the clarification! Didn't know about such a weird Firefox behavior before

oleksandr-danylchenko:
Seems like the Ponyfill introduces its own set of weird selection behaviors 😅 
https://github.com/recogito/text-annotator-js/issues/74#issuecomment-1979218606

rsimon:
It's safe to say that we won't be using the ponyfill anymore. I removed it manually (rather than resolving the conflicts that have come up recently.)

PR #34, opened 2024-02-12
#24 Added support for the omission of the not annotable elements within the text 

oleksandr-danylchenko:
## Relations
This PR is fully based on the https://github.com/recogito/text-annotator-js/pull/28 and https://github.com/recogito/text-annotator-js/pull/29 PRs. It includes all their changes as well. So I would recommend starting from them first.

## Issue
This PR resolves the https://github.com/recogito/text-annotator-js/issues/24 issue and helps to omit not annotable elements within the text.

## Demo
image

## Changes Made
In this PR I replaced the singular `selector: TextSelector;` with the plural `selector: TextSelector[];`. Each of those selectors matches the selection range on an annotable element!

A single selection range gets splitted when an element with the `data-not-annotable="true"` attribute is found within it. Also, the not annotable elements are not included in the position counting! It's convenient to skip elements with dynamic content.

## Further Steps
I didn't touch the TEI plugin yet, but it required Typescript alignment ❗


rsimon:
Thanks, I believe this looks good. But there's one thing we'll definitely need to address: the TextAnnotation type needs to provide some kind of legacy fallback. We have a substantial number of annotations in our production systems already, and need to keep supporting them. This means that a an annotation with a single TextSelector must still be valid.

rsimon:
P.S.: does the Range-reviving process take into account the non-annotatable areas? If so, I wonder if there's even a need to support multiple selectors. If the annotation is still defined by a single start and end point, it might be enough to stick with a single __TextSelector__ that contains an array of __Ranges__. Because the Ranges only exist on the client, and are not stored on the server, this wouldn't impact compatibility with existing annotations.

oleksandr-danylchenko:
> the TextAnnotation type needs to provide some kind of legacy fallback. We have a substantial number of annotations in our production systems already, and need to keep supporting them

That sounds like a reasonable concern! I suppose we can solve it in a few ways 🤔 
1. Converting the `TextAnnotation` interface to a `class`. It will execute crosswalk from a singular `selector: TextAnnotationTarget` into `selector: TextAnnotationTarget[]` within its `constructor`. And then all the calls to that property will work in the same, updated way.
2. Using [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) on the `TextAnnotation` objection. It will trap `get` and `set` operations for the `selector`. On `get` it'll always return an array, and on set it'll convert a singular assigned value to an array as well.

The 1st option seems a bit more straightforward to me. But it will require removing `TextAnnotation` _type_

oleksandr-danylchenko:
> P.S.: does the Range-reviving process take into account the non-annotatable areas? 

The `reviveRange` _takes_ the not annotable areas _into account_ by skipping them in the text nodes iterator: 
```ts
const iterator = document.createNodeIterator(offsetReference, NodeFilter.SHOW_TEXT, (node) =>
  node.parentElement?.closest(notAnnotableSelector)
    ? NodeFilter.FILTER_SKIP
    : NodeFilter.FILTER_ACCEPT
);
```
For example, we have 2 selectors with such start/end positions `[ [1, 50], [50, 100] ]`. The `reviveRange` on position 50 will retrieve a range that can actually start on position 350. But all those 300 characters were falling into a not annotable area and were skipped.

If we use just a single selector with a single pair of positions `[1, 100]` - there would be no way to know how we should split it into ranges w/o running a relatively heave `getAnnotableRanges` function

rsimon:
Hm, a part of me resists introducing a class for this... But I agree it might be the only way to avoid re-re-and-re-repeating `Array.isArray(...)` checks in 100 different places.

On second thought: in our production systems, we have, effectively, storage plugins that handle the communication with the backend. We could just add the crosswalk there. And then stick to an API that requires selector to be an array. Let me think about this for a while. But I believe I prefer this over a class.

oleksandr-danylchenko:
> Hm, a part of me resists introducing a class for this... But I agree it might be the only way to avoid re-re-and-re-repeating Array.isArray(...) checks in 100 different places.

Yep, exactly 👌🏻 
Using the `Array.isArray()...` all over the place will also create an implicit requirement for the developers to remember to add the check before using the prop 🥴 

--- 

> On second thought: in our production systems, we have, effectively, storage plugins that handle the communication with the backend. We could just add the crosswalk there.

That sounds like the best solution - to move the crosswalk out of the Core of the library 👍🏻

rsimon:
Re other issue:

> For example, we have 2 selectors with such start/end positions [ [1, 50], [50, 100] ]. The reviveRange on position 50 will retrieve a range that can actually start on position 350. But all those 300 characters were falling into a not annotable area and were skipped.

I should probably read the code first. But can you explain this again (like I'm 5 years old ;-)? So you're marking not-annotatable elements with the data attribute. And you evaluate this when you compute the selector from the selection range. Then when you re-generate an existing, your point is that running the same computation would be unnecessarily expensive, correct?

oleksandr-danylchenko:
> So you're marking not-annotatable elements with the data attribute. 

Yep. The `data-not-annotable` attribute can be set on the content element and its position is pre-defined by the text supplier. So we're not "making" elements not-annotable, but just "processing" those elements in a special way.

---

For beginning, let's break up the process of the selection that is handled by the `SelectionHandler`:
1. `onSelectStart` fires and a brand-new `target` is created with [`selector: []`](https://github.com/recogito/text-annotator-js/pull/34/files#diff-1e072053802159d91da89e331e1fe166da3cb395a3fc6ab3d6317b0510ea17e0R36).
2. `onSelectionChange` fires 
  2.1. It picks the 1st selection range from the `document.getSelection()`
  2.2. It [trims](https://github.com/recogito/text-annotator-js/pull/34/files#diff-1e072053802159d91da89e331e1fe166da3cb395a3fc6ab3d6317b0510ea17e0R57) that selection range (still a singular range!)
  2.3. The trimmed range i[s separated into N "annotable ranges"](https://github.com/recogito/text-annotator-js/pull/34/files#diff-1e072053802159d91da89e331e1fe166da3cb395a3fc6ab3d6317b0510ea17e0R58) by the [`getAnnotableRanges`](https://github.com/recogito/text-annotator-js/pull/34/files#diff-76ad7bd83653d8205d243c4b062b1c68526d99c0eef0213d5a98152421810c29R29) function
  2.4. It compares whether the `currentTarget.selector` content is somehow different from the newly obtained annotable ranges
  2.5. The `currentTarget` is updated and for each annotable range the [`rangeToSelector`](https://github.com/recogito/text-annotator-js/pull/34/files#diff-0b223530a99b27daba7084922f92e91a3fd31f5e2993714cac5d4058e4cc7c25R4-R8) function runs.
  That function creates the `{ quote, start, end }` object from a range. The logic for obtaining the `quote` and the `end` didn't change, only for the `start`. Previosly, we were creating a temp range from the start of the container to the range we're processing. And the `start` was equal to `rangeBefore.toString().length`. But that won't work adequately if the `rangeBefore` contains any not-annotable elements within! As their length can be dynamic and we wanna skip them from the position counting! So I made the `start` position to be obtained like `getRangeAnnotableContents(rangeBefore).textContent.lenght;`. [`getRangeAnnotableContents`](https://github.com/recogito/text-annotator-js/pull/34/files#diff-76ad7bd83653d8205d243c4b062b1c68526d99c0eef0213d5a98152421810c29R70-R74) just removes the elements with not annotable attribute from the range text.

Those were the steps how we _create_ those selectors! However, with that updated approach with skipping not annotable elements on `start` position obtaining wouldn't work for the old `reviveTarget` method! Because it was counting _all_ the characters, even from not-annotable elements to obtain the range! So I reworked that method as well 👌🏻 

Here's the interesting part in the new `reviveRange` function: 
```ts
const reviveRange = (start: number, end: number, offsetReference: HTMLElement): Range => {
  const iterator = document.createNodeIterator(offsetReference, NodeFilter.SHOW_TEXT, (node) =>
    node.parentElement?.closest(notAnnotableSelector)
      ? NodeFilter.FILTER_SKIP
      : NodeFilter.FILTER_ACCEPT
  );

  ...counting the characters to revive the range...
```
Now the revive method skips any text nodes are are located _within the not-annotable elements_! And the counting would match the numbers generated on the selection.

---

So let's get back to the example I made previously. 2 selectors generated on positions: `[ [1, 50], [50, 100] ]`. Such selectors for a single anno could be generated if within a selection range is a not-annotable element on position 50 - `[1, 50]`. The user went over that element and selected 50 more characters after it - `[50, 100]`.
(conceptual visual representation of such a selection)
image

oleksandr-danylchenko:
>  Then when you re-generate an existing selector, your point is that running the same computation would be unnecessarily expensive, correct?

For example, when the selectors are re-generated from the serialized list - then it's quite trivial to revive multiple ranges from the positions. `selectors [ [1, 50], [50, 100] ]` -> `ranges [ [1, 50], [350, 450] ]`.
But if we have just a single selector `[1, 100]` - it's quite burdensome to create those multiple ranges. We would need to split the range into N annotable ranges, ... (look at the point 2.3 and further)

rsimon:
Thanks for this explanation - that's very clear! 

Just two more question: 
- is there a specific need in your case to *really* treat the non-annotatable regions as if they don't exist at all? I.e. is their content likely to change? If it is - fully understood. If it isn't, you could still treat the annotation as a single range, and then just exclude the non-annotatable part in the graphical rendering step.
- as you have probably seen, there is also a `not-annotatable` CSS class, which causes selection start events to get ignored. I wonder if we could unify both "types" of markers? I realize they achieve slightly different things. (As a user, you might start your selection on a `data-not-annotatable` element, and that would be perfectly valid. The Annotator would just remove those bits from the selector.) But I believe it might be justifiable to have only one type of "not annotatable" marker, and then use it to control both behaviors. Might actually make more sense from the user perspective?

rsimon:
Re my first question: I revisited your original issue, and believe I already know the answer. (I guess you __do__ expect to have elements you might swap in and out of a page - like those interactive form elements - correct?)

rsimon:
And one last thing: given your use case, it still seems that your selectors would always look like this.

`[10, 100], [100, 150]` or `[12,123], [123,233]` etc.

I.e. you would __always__ be dealing with a continuous range, as far as the character count is concerned. I guess that still makes me feel a bit uneasy. The reviving step already does exclude the non-annotatable elements, therefore, the range `[10,150]` would revive start- and endpoints just as correctly as the two ranges `[10, 100]` and [100, 150], right? The difference would simply be that the Text Annotator would draw a highlight over the non-annotatable region. 

So, in short: I'm still wondering whether using only a single range + a bit of extra logic that excludes non-annotatable elements in the rendering step would be a cleaner solution.

oleksandr-danylchenko:
> is there a specific need in your case for treating the non-annotatable regions as if they really don't exist in the DOM at all? Are you expecting, for example, that their content will change (and that this change should not interfere with existing annotations)? I

Yep! That's the exact reason I even bother suggesting such a feature ;)
In our app, we have something like an interactive textbook. Among the textual content, users can answer various quizzes, polls, interact with the charts, etc. And such "interactive" content shouldn't be annotable. Because then we cannot reliably measure the position of the selectors that come after such elements with dynamic content

oleksandr-danylchenko:
> I.e. you would always be dealing with a continuous range, as far as the character count is concerned

I should have also specified that it won't **always** be the case. We intend to make the annotations resilient to the changes that can be introduced by the editors to the textual annotable content. So let's say that we are the same `[10, 100], [100, 150]` positions. At some point, an editor decides to insert another paragraph right before the not-annotable element at position 100. Technically, the previous 2nd range `[100, 150]` will now point to that new paragraph. But a user has never annotated it! To solve that issue, we're about to develop a rebase process on our side that will re-calculate the position of the 2nd range and will push the positions forward. That would make anno ranges look like: `[10, 100], [160, 210]`. And... here's the gap!
If we would use a single set of positions, such manipulations become super burdesome

rsimon:
Just quickly, to clarify: the not-annotatable CSS class does not prevent selecting (nor copy and paste)! Only the creation of annotations. Therefore I think there's only need for a single marker, right?

oleksandr-danylchenko:
> Just quickly, to clarify: the not-annotatable CSS class does not prevent selecting (nor copy and paste)! Only the creation of annotations. Therefore I think there's only need for a single marker, right?

Oh, then **yes**. We should be able to rely solely on the `.not-annotatable`  class 👍🏻

rsimon:
Ok, understood. However, the problem I see is that you're mixing two concerns. One is the ability to make parts of the DOM invisible to the annotator. The other is the ability to insert "break points" that allow you to change the content later.

I think I'm convinced that allowing multiple ranges per annotations is a good thing. (If for no other reason that annotations should be modeled after the browser selection API, and that allows multiple ranges, too.) 

But I'm not convinced that it's good to mix both concerns. You're essentially using the non-annotatable elements as a backdoor to splitting later on. If you never end up splitting annotations, it's an unnecessary feature. If you ever need to split at a location where you haven't put a non-annotatable element, your approach would work.

If the intention is to support mutable content, I believe a cleaner solution would be to go with some kind of XPath-based selector approach. Strategically placing non-annotatables instead feels like a hack TBH.

rsimon:
Ok, after giving this some more thought (and sleeping over it): you'll be happy to hear that I've come to the conclusion that the current approach is the best option after all :-) 

In the end, I felt that the Text Annotator data model should mimic the situation in the DOM as much as possible. Even if the character range is (by numbers) continuous, the DOM will use two `Range` objects. Therefore, we should also have two selectors. Otherwise, we'd again run into messy situations.

I still want to read through all of the code, to make sure I grok everything. But will try to merge the PR later today!

rsimon:
P.S.: I also want to address the issue of creating the annotation on pointer up. But since that's more of a performance optimization, let's wait for this until the other PRs/issues are resolved.

packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts
```diff
@@ -6,10 +6,9 @@ import {
   parseW3CUser,
   serializeW3CBodies
 } from '@annotorious/core';
-import type { TextAnnotation, TextSelector } from '../core';
+import type { TextAnnotation, TextAnnotationTarget, TextSelector } from '../core';
 import type { W3CTextAnnotation, W3CTextSelector } from '../w3c';
-import { reviveTarget } from '../../state';
-import { getQuoteContext } from '../../utils';
+import { getQuoteContext, reviveSelector } from '../../utils';
```
rsimon, line 11: A note on this: I don't think this should happen here. The core should handle range reviving, not the crosswalk. In fact I believe that's already the case... it should be safe to omit reviving the range here, since it will happen when the annotation gets added to the store. (But didn't test yet.)

oleksandr-danylchenko:
> Ok, after giving this some more thought (and sleeping over it): you'll be happy to hear that I've come to the conclusion that the current approach is the best option after all :-)

😅 😮‍💨😮‍💨😮‍💨

packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts
oleksandr-danylchenko: Unfortunately, I think removing the revival from the `parseW3CTextAnnotation` might was a mistake. Because:
1. It breaks the typescript contract. The `parseW3CTextAnnotation` returns the `TextAnnotation`, where the `range` is a mandatory property. By removing the revival - the parsed selector won't have it 🤷🏻‍♂️ 
2. It breaks the adapter's tests that expect the `range` to present on the parsed selector

oleksandr-danylchenko: I think that having 2 types might be a better path 🤔. Saying that the `range` is possibly undefined would be another TS "lie". As it cannot be the case within the Core. 

I would suggest adding not the `RevivedTextAnnotation`, but the `UnrevivedTextAnnotation` type instead. So we can leave all the existing code which uses the `TextAnnotation` in place. But precisely add usage of the `UnrevivedTextAnnotation` as a type for any annotations that come from outside of the lib

oleksandr-danylchenko: The timing for this discussion turned out to be just right 😆 
I've realized that for our app we'll need a custom `reviveSelector` function that will process positions differently. So if we use the `reviveSelector` in the W3C adapter - it would require rewriting the adapter as well in the app). But w/o it - we'll just monkey patch the `addAnnotation` in the plugin and will execute our custom revived there

rsimon: Sounds good! I agree about your first point, 2 types being the better solution. However, following suggestion: would it make sense to have a `TextAnnotation` type where the range is optional, and a `RevivedTextAnnotation` where range is mandatory?

In the "edges" of the API, it's often the case that someone will hand in one or the other. (An unrevived annotation loaded from the backend, or a revived annotation that the pulled out from a "create" event or whatever.) The `revive` methods don't really care. They will check if a (non-collapsed) range is there, and revive the range if there isn't. That means the revive methods would take any `TextAnnotation` as input, and always return a `RevivedTextAnnotation` as a result. 

This way, in the core, we'd always work with RevivedTextAnnotations, while users of the API generally wouldn't need to care. (RevivedTextAnnotation is also a TextAnnotation, after all.)

Even in your case, you _could_ write your own custom adapter that returns a RevivedTextAnnotation, yet still use the core unchanged, right? (Or you could monkey patch it, too. In fact I've done the same for the PDF annotator. Although that started to feel even messier than the rest of the text annotator ;-)

oleksandr-danylchenko: > would it make sense to have a TextAnnotation type where the range is optional, and a RevivedTextAnnotation where range is mandatory?

I would still recommend going the other way around. Because we already have soooo many places where the `TextAnnotation` is used. And renaming `TextAnnotation` to something else would lead to a pretty large hurdle. And we're not even talking about any code outside of the R6O library. So having another type with the optional `range` that will be used only on edges sound like a safer option to me

rsimon: I'll give it some thought. But TBH it just feels more sensible in terms of semantics to have the more generic name for the more generic "thing", while a more specialized name for the more narrowly defined type?

rsimon: Ok, I'm willing to go with the argument that `TextAnnotation` is the internal model, and that's supposed to come with range attached. But we'd definitely need a better name for the "other" thing then. "UnrevivedTextAnnotation" doesn't work, because the range would have to be optional, right? Something that conveys the idea that this could be both an internal annotation, as well as one that still needs to be revived.

oleksandr-danylchenko: `RangelessAnnotation`? 😆 

> Something that conveys the idea that this could be both an internal annotation, as well as one that still needs to be revived.

That's a valid concern. I see how `RevivedTextAnnotation` works now - it conveys the idea of "narrowing" down the type via "reviving". And there's no ambiguity in what the `Revived...` means 🤔 
But it's not obvious to think  of a good inverse name that would have the same directness level

PR #35, opened 2024-02-12
#24 Made the `TEI` extension compatible with the new `TextAnnotation`  with multiple selectors

oleksandr-danylchenko:
## Issue
This PR finishes the work started in #34 by making the `TEIAnnotationTarget` fully compatible with the [updated `TextAnnotationTarget`](https://github.com/recogito/text-annotator-js/pull/34/files#diff-817f7b7f19574595726de7163e819c9fcd3e4ce78df02d6cf307fd72dc62a374R11). So I would recommend reviewing the #34 first, merge it, and then look at this PR with a clear diff

## Changes Made
- `TEIAnnotationTarget` now uses the `selector: TEIRangeSelector[];` for compatibility with the `TextAnnotationTarget`. In that way `TEIPlugin` can reuse the `TextAnnotationStore` methods that depend on the `selector` array w/o any issues
- `crosswalk/forward` & `crosswalk/reverse` got the `textToTEITarget`/`teiToTextTarget` updated to support an array of selectors


rsimon:
I made a few refactorings + tweaks to ensure that the TEI extension and the React binding keep working. I think this PR is no longer necessary now?

oleksandr-danylchenko:
Thanks for the cleanup, it looks great 🌟 
I agree that this PR is not needed anymore, as the build for the TEI plugin passes 👌🏻

Issue #37, opened 2024-02-19
Passing `container` to the `W3CFormat` using the React `TextAnnotator`


oleksandr-danylchenko:
## Issue
For the `W3CFormat` to work, we need to provide the `container `argument: 
https://github.com/recogito/text-annotator-js/blob/9dbac6590ed5c2936c7bb28a5739b8f06209d3f3/packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts#L20-L26

It's pretty trivial when we're using plain JS for the app, as we can just query the container: 
https://github.com/recogito/text-annotator-js/blob/9dbac6590ed5c2936c7bb28a5739b8f06209d3f3/packages/text-annotator/test/index.html#L287-L291

---

But for the React... that container is being automatically created and concealed within the `TextAnnotator` ...
https://github.com/recogito/text-annotator-js/blob/9dbac6590ed5c2936c7bb28a5739b8f06209d3f3/packages/text-annotator-react/src/TextAnnotator.tsx#L48-L52

So when you'd wanna to create the `W3CFormat` and pass it to the `TextAnnotator` as a prop there's no obvious way to do that:
![image](https://github.com/recogito/text-annotator-js/assets/68850090/944c6726-aa5d-49f5-a85a-d40a609505da)

## Possible Solutions
Make the `TextAnnotator` use the `forwardRef` to gain access to the automatically created container via the `ref`. Pretty straightforward 👌🏻 
```tsx
export const TextAnnotator = forwardRef<HTMLDivElement>(<E extends unknown>(props: TextAnnotatorProps<E>, ref) => {
```

PR #38, opened 2024-02-19
#37 Added ref forwarding for the `TextAnnotator` component

oleksandr-danylchenko:
## Issue
This PR resolves the #37 issue and helps to access the content container generated within the `TextAnnotator` component

So now we can create the `W3CFormat` adapter on the consuming app's side:
image


packages/text-annotator-react/src/TextAnnotator.tsx
```diff
@@ -7,26 +8,27 @@ import '@recogito/text-annotator/dist/text-annotator.css';
...
+  children?: ReactNode | JSX.Element;
 
   filter?: Filter;
 
   style?: DrawingStyle | ((annotation: TextAnnotation) => DrawingStyle);
 
 }
 
-export const TextAnnotator = <E extends unknown>(props: TextAnnotatorProps<E>) => {
+export const TextAnnotator = forwardRef(<E extends unknown>(props: TextAnnotatorProps<E>, ref: ForwardedRef<HTMLDivElement>) => {
```
oleksandr-danylchenko, line 19: I didn't use the generic type of the `forwardRef`, because it would allow using the `<E extends unknown>` for the `TextAnnotatorProps`

packages/text-annotator-react/package.json
```diff
@@ -41,6 +41,7 @@
   },
   "dependencies": {
     "@neodrag/react": "^2.0.3",
-    "CETEIcean": "^1.9.2"
+    "CETEIcean": "^1.9.2",
+    "react-merge-refs": "^2.1.1"
```
oleksandr-danylchenko, line 45: I decided that it would be more convenient to use a small and maintained helper package than to add the utility manually 🤷🏻‍♂️

PR #40, opened 2024-02-19
#37 Added callback `adapter` for the  React `TextAnnotator`

oleksandr-danylchenko:
## Issue
This PR resolves the #37 issue and makes it possible to provide the `W3CFormat` adapter into the `TextAnnotator` component

## Changes Made
I implemented the 2nd approach suggested in the #37. It allowed having a simple component's API on the consumer side and not dealing with the `undefined` container element manually.


packages/text-annotator-react/src/TextAnnotator.tsx
```diff
@@ -1,13 +1,16 @@
 import { ReactNode, useContext, useEffect, useRef } from 'react';
 import { AnnotoriousContext, DrawingStyle, Filter } from '@annotorious/react';
+import type { FormatAdapter } from '@annotorious/core';
 import type { TextAnnotation, TextAnnotatorOptions } from '@recogito/text-annotator';
 import { createTextAnnotator } from '@recogito/text-annotator';
 
 import '@recogito/text-annotator/dist/text-annotator.css';
 
-export type TextAnnotatorProps<E extends unknown> = TextAnnotatorOptions<E> & {
+export interface TextAnnotatorProps<E extends unknown> extends Omit<TextAnnotatorOptions<E>, 'adapter'>  {
```
oleksandr-danylchenko, line 9: W/o using the `interfaced`  Typescript still points to the `adapter` prop from the `TextAnnotatorOptions`, not the overridden one

packages/text-annotator-react/src/TextAnnotator.tsx
```diff
@@ -26,7 +29,9 @@ export const TextAnnotator = <E extends unknown>(props: TextAnnotatorProps<E>) =
   useEffect(() => {
     if (!setAnno) return;
 
-    const anno = createTextAnnotator(el.current, opts);
+    const adapter = typeof opts.adapter === 'function' ? opts.adapter(el.current) : opts.adapter;
```
oleksandr-danylchenko, line 32: I'm not using a more laconic `opts` overriding
```ts
opts.adapter = typeof opts.adapter === 'function' ? opts.adapter(el.current) : opts.adapter;
```
because it would mutate the props

packages/text-annotator-react/test/App.tsx
```diff
@@ -62,10 +62,11 @@ const MockStorage = () => {
 }
 
 export const App = () => {
-  
+  const w3cAdapter = useCallback((container: HTMLElement) => W3CTextFormat('https://www.gutenberg.org', container), []);
+
   return (
     <Annotorious>
-      <TextAnnotator>
+      <TextAnnotator adapter={w3cAdapter}>
```
oleksandr-danylchenko, line 69: Simple one-liner definition for the formatter on the consumer side, where the `container` is always defined

packages/text-annotator-react/src/TextAnnotator.tsx
```diff
@@ -1,13 +1,16 @@
...
 
 import '@recogito/text-annotator/dist/text-annotator.css';
 
-export type TextAnnotatorProps<E extends unknown> = TextAnnotatorOptions<E> & {
+export interface TextAnnotatorProps<E extends unknown> extends Omit<TextAnnotatorOptions<E>, 'adapter'>  {
 
-  children?: ReactNode | JSX.Element; 
+  children?: ReactNode | JSX.Element;
+
+  adapter?: FormatAdapter<TextAnnotation, E> | ((container: HTMLElement) => FormatAdapter<TextAnnotation, E>) | null;
```
oleksandr-danylchenko, line 13: 1. W/o using the `interfaced`  Typescript still points to the `adapter` prop from the `TextAnnotatorOptions`, not the overridden one
2. `adapter` is overridden only within the `TextAnnotator.tsx`, because only the React wrapper automatically creates a container element:
https://github.com/recogito/text-annotator-js/blob/9dbac6590ed5c2936c7bb28a5739b8f06209d3f3/packages/text-annotator-react/src/TextAnnotator.tsx#L48-L52

rsimon:
Thanks, I'll merge this! 

I've been thinking about this myself. Although I hadn't thought as far as the React wrapper yet (so - thanks for this!) I was wondering whether we could somehow reorganize things, simply to make the API look nicer. I.e. get from the current...

```
var r = createTextAnnotator(contentContainer, {
  adapter: W3CTextFormat('https://www.gutenberg.org', contentContainer)
});
```

...to something like this again:

```
var r = createTextAnnotator(contentContainer, {
  adapter: W3CTextFormat('https://www.gutenberg.org')
});
```

I guess, with your current PR, we're almost there. `W3CTextFormat('...')`  could itself return a function(container). In this case, we'd get the same behavior in React __and__ a leaner API in vanilla JS. (Not sure how that would play with callback and re-rendering in React tho?)

oleksandr-danylchenko:
> W3CTextFormat('...') could itself return a function(container). In this case, we'd get the same behavior in React and a leaner API in vanilla JS.

That's an interesting suggestion 👍🏻 
I may get back on this in some time when I'll finish with project-specific work

Issue #41, opened 2024-02-19
Missing `selected` argument in the `style` property callback


oleksandr-danylchenko:
## Issue
I noticed inconsistent definitions for the `style` property across the packages. But the main inconveniences are:
- public types defines `style` callback w/o providing a selection state as the 2nd argument
https://github.com/recogito/text-annotator-js/blob/c9bff1f651ad06de7cd3c86d0e429ae1a64c1497/packages/text-annotator/src/TextAnnotatorOptions.ts#L17
Although, such an argument is being passed from within the `highlightRenderer`:
https://github.com/recogito/text-annotator-js/blob/c9bff1f651ad06de7cd3c86d0e429ae1a64c1497/packages/text-annotator/src/highlight/canvas/highlightRenderer.ts#L95-L98
- `style` prop uses an A9S-specific `DrawingStyle` although there's the `HighlightStyle` type present. The latter contains only the props that can be processed by the `text-annotator-js`:
https://github.com/recogito/text-annotator-js/blob/c9bff1f651ad06de7cd3c86d0e429ae1a64c1497/packages/text-annotator/src/highlight/HighlightStyle.ts#L13-L23

## Suggested Changes
Create a _shared_ type definition for the `style` prop. It'll also utilize the new `HighlightStyle` type, instead of the `DrawingStyle`. Also, it would be _preferred_ to make the 2nd argument a _broader object_. In this way, we'll be able to easily extend it with info about hovering, etc.
```ts
export type HighlightStyleOption = HighlightStyle | ((annotation: Annotation, state: { selected: boolean }) => HighlightStyle) | undefined;
```

oleksandr-danylchenko:
Closing it due to the upcoming changes for the `HighlightStyleExpression`: https://github.com/recogito/text-annotator-js/pull/42#issuecomment-2022599303

PR #42, opened 2024-02-19
#41 Unified `style` prop with `isSelected` state argument

oleksandr-danylchenko:
## Issue
This PR resolves the #41 issue when the public types defined the `style` callback w/o providing a selection state as the 2nd argument.

## Changes Made
- Renamed `HighlightStyle` -> `HighlightDrawingStyle`. The `HighlightDrawingStyle` partially extends the A9S `DrawingStyle` and narrows it down to the text annotation needs. Therefore, it would be appropriate to express that connection in the name, IMHO
- Added a shared aggregated type with the `HighlightStyle` name:
```ts
export type HighlightStyle =
  | HighlightDrawingStyle
  | ((annotation: TextAnnotation, isSelected?: boolean) => HighlightDrawingStyle)
  | undefined;
```
It's a part of the public API that uses the newer `HighlightDrawingStyle` and also exposes the `isSelected` flag. Using that single type helps with deduping and adds an ability to extend the exposed state in a single place

## Further Changes
The single `isSelected` argument can be changed to a `state: { selected: boolean }` object. In that way, adding new flags would become much easier

### DIFF
--- a/packages/text-annotator/src/highlight/index.ts
+++ b/packages/text-annotator/src/highlight/index.ts
@@ -1,2 +1,4 @@
+export * from './HighlightPainter';
+export * from './HighlightStyle';
 export * from './canvas';
 export * from './css';
--- a/packages/text-annotator-react/src/TextAnnotator.tsx
+++ b/packages/text-annotator-react/src/TextAnnotator.tsx
@@ -1,6 +1,6 @@
 import { ReactNode, useContext, useEffect, useRef } from 'react';
 import { AnnotoriousContext, Filter } from '@annotorious/react';
-import type { HighlightStyleOption, TextAnnotatorOptions } from '@recogito/text-annotator';
+import type { HighlightStyle, TextAnnotatorOptions } from '@recogito/text-annotator';
 import { createTextAnnotator } from '@recogito/text-annotator';
 
 import '@recogito/text-annotator/dist/text-annotator.css';
@@ -11,7 +11,7 @@ export type TextAnnotatorProps<E extends unknown> = TextAnnotatorOptions<E> & {
 
   filter?: Filter;
 
-  style?: HighlightStyleOption;
+  style?: HighlightStyle;
 
 }
 
--- a/packages/text-annotator-react/src/TextAnnotator.tsx
+++ b/packages/text-annotator-react/src/TextAnnotator.tsx
@@ -1,6 +1,6 @@
 import { ReactNode, useContext, useEffect, useRef } from 'react';
-import { AnnotoriousContext, Filter } from '@annotorious/react';
-import type { HighlightStyle, TextAnnotatorOptions } from '@recogito/text-annotator';
+import { AnnotoriousContext, Filter, FormatAdapter } from '@annotorious/react';
+import type { HighlightStyle, TextAnnotation, TextAnnotatorOptions } from '@recogito/text-annotator';
 import { createTextAnnotator } from '@recogito/text-annotator';
 
 import '@recogito/text-annotator/dist/text-annotator.css';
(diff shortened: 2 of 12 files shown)


packages/text-annotator/src/highlight/index.ts
```diff
@@ -1,2 +1,4 @@
+export * from './HighlightPainter';
+export * from './HighlightStyle';
```
oleksandr-danylchenko, line 2: I think that those files also should be re-exported because they are part of a public API and will be utilized by users who want to create custom painters or extend the styles with their app-specific props

rsimon:
+1 on `HighlightDrawingStyle`.

Also +1 on the introduction of a shorthand for `HighlightDrawingStyle | ((annotation: TextAnnotation, isSelected?: boolean) => HighlightDrawingStyle) | undefined;` But I'm not sure I like `HighlightStyle` as a name. IMO too close to `HighlightDrawingStyle`. How about `HighlightDrawingExpression` (courtesy of ChatGPT).

Maybe to shorten, we can get rid of the 'Drawing', and we might not need the `| undefined` because we can cover that with `?` for an optional param?

That would result in:

```
export interface HighlightStyle extends Pick<DrawingStyle, 'fill' | 'fillOpacity'> {

  underlineStyle?: string;

  underlineColor?: number;

  underlineOffset?: number;

  underlineThickness?: number;
  
}

export type HighlightStyleExpression = 
  HighlightStyle | ((annotation: TextAnnotation, isSelected?: boolean) => HighlightStyle);
```

oleksandr-danylchenko:
> and we might not need the | undefined

I think so. It might be a better types API design to let the functions control whether the type is required or not 👍🏻 

> `export interface HighlightStyle extends Pick<DrawingStyle, 'fill' | 'fillOpacity'> {`

I would like to _keep the `HighlightDrawingStyle` name_ to express its connection with the A9S `DrawingStyle` 🤔

oleksandr-danylchenko:
> But I'm not sure I like HighlightStyle as a name. IMO too close to HighlightDrawingStyle. How about HighlightDrawingStyleExpression

That's an option... But in my view (possibly author-biased), those names are not _that_ close... Because the `HighlightDrawingStyle` describes a flat object with properties used to draw the highlight. But the `HighlightStyle` is a generic name describing how the `HighlightDrawingStyle` can be passed anywhere. It was intentionally made shorter for better convenience, using it in function signatures.
But I'm not yet 100% locked on that specific name yet... I need to give it a thought 🤔

rsimon:
I've been considering 'config' myself, but then didn't really like it on 2nd thought. Because it's really not a configuration. ChatGPT suggested "expression" as an umbrella term for something that could be either a _value_, or a _function that returns that value_, and I tend to like it TBH.

oleksandr-danylchenko:
> as an umbrella term for something that could be either a value, or a function that returns that value

Interesting... I initially thought about the `Expression` as yet another overly generic term like a "manager". But turned out I just forgot its original semantical meaning.
After I looked it up again - I think the `HighlightStyleExpression` fits us perfectly!

rsimon:
Hi, just an quick heads-up that I will close this PR. I'm in the midst of refactoring some things in the core. This will almost have conflicts later today. But just to say: both the 'HighlightStyleExpression' and the change from the `isSelected` boolean to a proper annotation state object are now onboard 👍

rsimon:
P.S.: there might be a few small breaking changes. But hopefully nothing problematic. (I've mostly cleaned up the naming on some of the store methods.)

oleksandr-danylchenko:
> But just to say: both the 'HighlightStyleExpression' and the change from the isSelected boolean to a proper annotation state object are now onboard 👍

Great, thanks for the update!

PR #44, opened 2024-02-19
#41 Made the `painter` and the `style` props receive the highlight state object

oleksandr-danylchenko:
## Issue
This PR extends the work done in #42 PR by replacing a single `isSelected?: boolean` flag with the `state?: { selected?: boolean }` object. That allows to add new state props like `hovered`, `focused`, etc. easier in the future w/o bloating the arguments list.
This PR is optional and somehow based on [this message](https://github.com/recogito/text-annotator-js/pull/19#issuecomment-1907897420):
> The second argument will change. It's currently just a boolean flag (isSelected). But will turn into an object which includes the following:
> 
> * selection (boolean)
> * mouse hover (boolean)
> * custom (an object that can be manipulated by the application - this is to support use cases where, e.g., an application has a list of annotations outside the text/image, and wants to style annotations according to whether a user selects them in the list.)


rsimon:
Like PR #42, this will no longer fit without conflicts after the upcoming refactoring. But  the key idea of an annotation state  will be onboard.

PR #45, opened 2024-02-19
Minor fixes + library upgraded

rsimon:


packages/text-annotator/src/utils/reviveTarget.ts
```diff
@@ -7,7 +7,7 @@ export const reviveTarget = (target: TextAnnotationTarget, container: HTMLElemen
     ? target
     : ({
       ...target,
-      selector: target.selector.map(s => s.range instanceof Range ? s : reviveSelector(s, container))
+      selector: target.selector.map(s => s.range instanceof Range && !s.range.collapsed ? s : reviveSelector(s, container))
```
oleksandr-danylchenko, line 10: `isRevived` can be used here

PR #59, opened 2024-02-26
Renamed `rangeToTEIRangeSelector` -> `textToTEISelector`

oleksandr-danylchenko:
When I was working on the plugin for my app - I faced a need to transform the `TextSelector` to some custom selector type.
I noticed that `TEI` plugin also does that. But the name for the function is a bit unintuitive 🤔 
It's called `rangeToTEIRangeSelector` although it receives the `selector` and converts to a more specific selector. Shouldn't it be named `selectorToTEIRangeSelector`? 🤔

### DIFF
--- a/packages/extension-tei/src/crosswalk/forward.ts
+++ b/packages/extension-tei/src/crosswalk/forward.ts
@@ -84,7 +84,7 @@ const toTEIXPaths = (startPath: string[], endPath: string[], selectedRange: Rang
  * Using the DOM Range from a (revived!) TextSelector, this function computes
  * the TEIRangeSelector corresponding to that range.
  */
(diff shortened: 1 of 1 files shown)


rsimon:
I agree the naming needs to be cleared up. But I think this one's not quite right either. The method converts a normal TextSelector to a TEI [XPATH Range Selector](https://www.w3.org/TR/annotation-model/#range-selector). So I guess the proper name would be  `textToTEIRangeSelector` or perhaps even `textToTEISelector`, because the "Range" (which is a WebAnno selector range, rather than a DOM Range just adds confusion?)

oleksandr-danylchenko:
> textToTEISelector

I like this one 👍🏻 
Also it matches the naming for the following `textToTEITarget` & `textToTEIAnnotation`

Issue #47, opened 2024-02-20
Allow using the `Annotorious` hooks with the `TextAnnotator` model


oleksandr-danylchenko:
## Current State
The `Annotorious.tsx` exports a few [public hooks](https://github.com/annotorious/annotorious/blob/8de692434213ad910df3af53fef28b620c60ffc9/packages/annotorious-react/src/Annotorious.tsx#L111-L193) that utilize the shared [`AnnotoriousContext`](https://github.com/annotorious/annotorious/blob/8de692434213ad910df3af53fef28b620c60ffc9/packages/annotorious-react/src/Annotorious.tsx#L99-L106). Unfortunately, all those hooks, except the `useAnnotator`, are tightly bound to the _A9S'es `Annotation` model_:
![image](https://github.com/recogito/text-annotator-js/assets/68850090/689189a7-da5e-4699-aebb-5ae596656f6f)

## Issues
1. When you try using `useAnnotationStore` hook with the `text-annotator-js` type `TextAnnotationStore` - you'll have a TS error!
![image](https://github.com/recogito/text-annotator-js/assets/68850090/7162c4c1-268d-4909-8783-013170d31b53)
That would be expected because the `TextAnnotationStore` doesn't fully extend the `Store` type:
https://github.com/recogito/text-annotator-js/blob/bd9e75cd682b809c54c46a3b78bf8bcaeff4a3b2/packages/text-annotator/src/state/TextAnnotationStore.ts#L4-L12
Unfortunately, it blocks users from accessing the Core's store with typing relevant to the `TextAnnotation` 🤷🏻‍♂️ 

2. The `TextAnnotator` doesn't specify the type of the `state` explicitly, but just uses the inherited one:
https://github.com/recogito/text-annotator-js/blob/bd9e75cd682b809c54c46a3b78bf8bcaeff4a3b2/packages/text-annotator/src/TextAnnotator.ts#L13-L20
Therefore, the type of the `annotator.state.store` would be `Store<TextAnnotation>` instead of the expected `TextAnnotatorStore` 🤷🏻‍♂️ 

## Possible Solutions

1. Change the `useAnnotationStore` type to the `const useAnnotationStore = <T extends unknown =  Store<Annotation>>() => {`
[...]

PR #60, opened 2024-02-26
#47 Added missing required `origin` for `addAnnotation` call

oleksandr-danylchenko:
## Issue
After in #48 PR I added usage of the actual `TextAnnotatorStore` - the `store.addAnnotation` within the `SelectionHandler` started to fail. That's because the store defines the `origin` as a mandatory argument:

https://github.com/recogito/text-annotator-js/blob/3f587dd752084f3f6c291dec0577e01afb2d1e4d/packages/text-annotator/src/state/TextAnnotationStore.ts#L8

### DIFF
--- a/packages/extension-tei/src/crosswalk/forward.ts
+++ b/packages/extension-tei/src/crosswalk/forward.ts
@@ -110,7 +110,7 @@ export const rangeToTEIRangeSelector = (selector: TextSelector): TEIRangeSelecto
   };
 }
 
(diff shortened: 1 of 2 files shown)


rsimon:
Ah - good catch! But the wrong fix ;-) Origin shouldn't be required. I.e. it's actually the Store type that should be fixed.

```
addAnnotation(annotation: TextAnnotation, origin?: Origin): boolean; 
```

Feel free to either make the change, or close the PR & I can make it.

oleksandr-danylchenko:
No problem, I can add the change to the store 👌🏻 
Thanks for the update

PR #61, opened 2024-02-26
#47 Made `origin` optional for the `TextAnnotationStore` methods

oleksandr-danylchenko:
## Issue
After in #48 PR I added usage of the actual `TextAnnotatorStore` - the `store.addAnnotation` within the `SelectionHandler` started to fail. That's because the store defines the `origin` as a mandatory argument:
https://github.com/recogito/text-annotator-js/blob/3f587dd752084f3f6c291dec0577e01afb2d1e4d/packages/text-annotator/src/state/TextAnnotationStore.ts#L8

However, according to @rsimon, having a mandatory `origin` for the `TextAnnotatorStore` methods is redundant


oleksandr-danylchenko:
Opps... Wrong branch pushed

Issue #66, opened 2024-02-27
Do we actually need the `trimRange` method? 🤔


oleksandr-danylchenko:
## TL;DR;
I think that we don't need to trim the selection range because:
1. User can intentionally _select non-textual nodes_. Taking them out of the selection range _only_ when they are at the edges, but not in the middle, seems like a partial prevention of a behavior
2. Ranges that have the `start/endContainer.nodeType` different from the `Node.TEXT_NODE` still can contain only the textual annotatable content. And such ranges are usually created when using the `.setStartAfter`/`. setEndBefore`

## Issue
In the https://github.com/recogito/text-annotator-js/issues/65 I brought up an issue with the `trimRange` method which is not that universal and cannot handle the trimming of the ranges that were split on `.not-annotatable` elements.
Then, in  https://github.com/recogito/text-annotator-js/issues/23 I noticed that `trimRange` doesn't act adequately as soon as the selection ends on a non-textual element: https://github.com/recogito/text-annotator-js/issues/23#issuecomment-1966301525.

---

All that led to the next **_conceptual questions about the purpose of the `trimRange` and whether we need it 🤔_**.

> The purpose of the method is to handle edge cases where the selection starts/ends on all sorts of non-text (images, layout block element, etc.) nodes. ©https://github.com/recogito/text-annotator-js/issues/23#issuecomment-1916179300

So the trim will discard the non-text nodes _only_ if they are _at the edges_ of the selection... but not in the middle... Although in both cases user _selected those nodes_.

For example, if I select the `text` + `img`, the latter will get removed. But if the selection is `text` + `img` + `text` - the image is kept in! Then why was it discarded in the first place 🤷🏻‍♂️. The same can be said for any other non-textual nodes that are trimmed on the edges, but not excluded from the middle of the selection. 
I feel like the not annotatable concept and the "selection with holes" fit such elements more 🤔 

---

Following that logic, I'm not sure about the robustness of the next checks we have in the code:
https://github.com/recogito/text-annotator-js/blob/3a70b9ef5cafec4b54ed629e934e5875d6f06262/packages/text-annotator/src/utils/trimRange.ts#L4-L9
https://github.com/recogito/text-annotator-js/blob/3a70b9ef5cafec4b54ed629e934e5875d6f06262/packages/text-annotator/src/state/spatialTree.ts#L42-L49
[...]

Issue #23, opened 2024-01-29
Question about `trimRange` for non-textual `startContainer` handling 🤔


oleksandr-danylchenko:
In the method `trimRange` there's a handling for the cases when either the `startContainer` or the `endContainer` are not the `Node.TEXT_NODE` type:
https://github.com/recogito/text-annotator-js/blob/80c2f9235ef27b2e57ff84a23d1f100a7b377e83/packages/text-annotator/src/utils/trimRange.ts#L11-L31
And on [line 19](https://github.com/recogito/text-annotator-js/blob/80c2f9235ef27b2e57ff84a23d1f100a7b377e83/packages/text-annotator/src/utils/trimRange.ts#L19) I noticed a small "weirdness". The comments say we're looking for the _first_ text child. 
**_Shouldn't we use `range.setStart` and not the `setEnd`?_** Am I missing something?🤔

rsimon:
Good catch. Yes, I think you are correct - that might be an error on my part. Would need testing though... The purpose of the method is to handle edge cases where the selection starts/end on all sorts of non-text (images, layout block element etc.) nodes. Doesn't often happen if the annotated document is mostly text. But happens if markup is more complicated, and users click around frantically (which the usually do ;-) I'll try to test this on some of our own more complex docs.

oleksandr-danylchenko:
### I think I noticed a few issues with the current `trimRange` method... 

1. In #65 I noticed that the `trimRange` [cannot be applied](https://github.com/recogito/text-annotator-js/issues/65#issuecomment-1966144252) to the ranges that were created after spitting on `.not-annotatable` elements
2. It creates weird collapsed ranges when the selection ends on an image:
![image](https://github.com/recogito/text-annotator-js/assets/68850090/13a0fc0f-d03d-4018-bb75-717398b9b650)
W/o the `trimRange` - the selector is correct!
![image](https://github.com/recogito/text-annotator-js/assets/68850090/79c1ebe2-2cb8-4773-8596-1f50a65967b5)

oleksandr-danylchenko:
And then there's a conceptual question about the role of the `trimRange` in general 🤔 
> The purpose of the method is to handle edge cases where the selection starts/ends on all sorts of non-text (images, layout block element etc.) nodes.

So the trim will discard the non-text nodes only if they are _at the edges_ of the selection... but not in the middle... Although in both cases user _selected those nodes_.

[...]

PR #67, opened 2024-02-27
#23 Removed `trimRange` usage

oleksandr-danylchenko:
## Issue
This PR covers the issue #66. Where I expressed my doubts about whether we need a `trimRange` in the selection range processing flow. Two main points that shifted me towards the "No" answer are: 

> 1. User can intentionally _select non-textual nodes_. Taking them out of the selection range _only_ when they are at the edges, but not in the middle, seems like a partial prevention of a behavior
> 2. Ranges that have the `start/endContainer.nodeType` different from the `Node.TEXT_NODE` still can contain only the textual annotatable content. And such ranges are usually created when using the `.setStartAfter`/`. setEndBefore`

But to grasp the whole picture you can get acquainted with the detailed description and thought process here - https://github.com/recogito/text-annotator-js/issues/66#issue-2156559097


oleksandr-danylchenko:
The only I noticed in the React after removing the `trimRange` is that the "empty" ranges between the elements are captured: 

https://github.com/recogito/text-annotator-js/assets/68850090/06589b94-8db0-4c81-b9e3-e5c9583eb881

When I was using the `trimRange` - it was throwing an exception, stopping the selection flow:

https://github.com/recogito/text-annotator-js/assets/68850090/4405ba81-249b-4adf-b51c-10f476bc4cb5

rsimon:
The `trimRange` is really a remnant of the [previous version of the text annotator](https://github.com/recogito/recogito-js) or, in fact, even the [previous version's predecessor](https://github.com/pelagios/recogito2/blob/main/app/assets/javascripts/document/annotation/text/selection/selectionHandler.js) (which worked in a "controlled environment" of a single DIV). Back then, the goal was simply to remove trailing and leading spaces.

I'll review this PR when I have the time, but I believe `trimRange` is indeed not needed any more.

oleksandr-danylchenko:
Thank you very much for the feedback 🙏🏻 
The doubts of me just using the library in some unintended way were tearing me apart. And submitting such a PR would be considered a super bold move 😅

Issue #68, opened 2024-02-27
Fix capturing invisible annotations on the "empty" textual nodes


oleksandr-danylchenko:
## Issue
When a user makes a selection between the paragraphs an "empty" annotation gets created. But it captures a bit different content based on the type of app you're working with:

### HTML pages, [like this](https://github.com/recogito/text-annotator-js/blob/main/packages/text-annotator/test/index.html)

https://github.com/recogito/text-annotator-js/assets/68850090/27753b4c-730a-4d6f-9640-e0cbbae19831

When selecting smth on a plain HTML page - the newline symbols and whitespaces are also captured in the range: 
![image](https://github.com/recogito/text-annotator-js/assets/68850090/f466a61e-7999-4999-b9fc-24cd54cf065b)
_Therefore, even if the range's content doesn't contain any actual characters - it still can be captured and annotated. But that annotation won't be visible to anyone 😅_ 

### React pages, [like this](https://github.com/recogito/text-annotator-js/blob/main/packages/text-annotator-react/test/index.html)

https://github.com/recogito/text-annotator-js/assets/68850090/8969e176-d9fb-46aa-8557-b928755e8108

Fortunately, this issue _isn't that critical for the React-based pages_ 🎉. Because it automatically wipes out non-textual symbols between the elements. So despite the captured range not being collapsed - the `hasChanged` check will prevent the creation of the invisible annotation:
![image](https://github.com/recogito/text-annotator-js/assets/68850090/6c7dbfe2-f290-438f-a5b4-da4191d6dbe3)
 https://github.com/recogito/text-annotator-js/blob/3a70b9ef5cafec4b54ed629e934e5875d6f06262/packages/text-annotator/src/SelectionHandler.ts#L56-L63

## Possible Solution
Manually verify that the content of the selected range contains meaningful annotatable content, and not just formatting characters like the `\n` & ` `. After that - those ranges can be discarded to prevent passing of the `hasChanged` check

rsimon:
This is now merged :-)

oleksandr-danylchenko:
Thank you 🚀

PR #69, opened 2024-02-27
#68 Fixed capturing invisible annotations on "empty" nodes

oleksandr-danylchenko:
## Issue
This PR resolves the #68 issue. Now the range won't be processed if it only consists of the whitespace characters or it's empty. It'll help to avoid the unnecessary creation of invisible annotations.
More details and examples can be found on the issue page - #68


oleksandr-danylchenko:
Could this PR be merged to reduce the number of opened ones? 🙏🏻 
It works pretty reliably in my app and I'm not getting phantom selections anymore

rsimon:
Sorry - needed to find the time to look into this, but: looks good, thanks!

Issue #71, opened 2024-02-29
Fix losing the `target.styleClass` property on the `W3CTextAnnotation` parsing


oleksandr-danylchenko:
## Issue
When a user chooses a highlighting color - that piece of data should be recorded into the [`styleClass` prop](https://arc.net/l/quote/uvmonnyi).

In my app, I manually extended the `TextAnnotationTarget` type in a `.d.ts` file with `styleClass`. So I'm able to record it with `store.updateAnnotation(merge({}, annotation, { target: { styleClass } }))`. Which gets correctly serialized for the W3C annos:
https://github.com/recogito/text-annotator-js/blob/3a70b9ef5cafec4b54ed629e934e5875d6f06262/packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts#L141-L146

But, unfortunately, not having that type on the `TextAnnotation` itself - led to a parsing bug:
https://github.com/recogito/text-annotator-js/blob/3a70b9ef5cafec4b54ed629e934e5875d6f06262/packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts#L42-L48
_The `styleClass` from the w3c anno never ends up on the core model and we lose it... 😭_
![image](https://github.com/recogito/text-annotator-js/assets/68850090/fe39bcd4-1976-413c-b0f9-9d0a7984f8c1)

---

The question of whether we should have the `styleClass` on the core model has already [been brought up previously](https://github.com/recogito/text-annotator-js/pull/17#issuecomment-1904022639). The decision at that point was to not add it because:
> it should stay out of the core. In the longer run (hopefully not too far future...) both the `selectAction` and the `style` function would be affected by the adapter. Which means you would get a `W3CTextAnnotation` as an argument to your `style` (or `pointerAction`) function.

Unfortunately, now I see a small flaw in that logic... If we don't have the `styleClass` somewhere in the Core model - there will nothing for the adapter to put into the `W3CTextAnnotation` 🤷🏻‍♂️ 

## Possible Solution
1. Add the `styleClass?: string` type to the `TextAnnotation` model
2. Add copying of the `styleClass` from the `W3CTextAnnotation` to the `TextAnnotation`

PR #72, opened 2024-02-29
#71 Fixed losing the `styleClass` on `W3CTextAnnotation` parsing

oleksandr-danylchenko:
## Issue
This PR resolves the #71 issue. Left there a detailed explanation behind the change

## Changes Made
1. Added copying of the `styleClass` from the `W3CTextAnnotation` to the `TextAnnotation`
3. Added error reporting when there are no targets on the annotation


oleksandr-danylchenko:
Could this PR be merged to reduce the number of opened ones? 🙏🏻

oleksandr-danylchenko:
Merge candidate? Seems like a pretty small, but beneficial, change 🤔

oleksandr-danylchenko:
@rsimon, I reworked the PR according to the approach you took for the `scope` property in https://github.com/recogito/text-annotator-js/commit/068e537c97b563c7af25f6141a13b8433d77cbdd.
Now the `styleClass` is "silently" serialized/deserialized in the core model and I used milder `@ts-expect-error` instead of `@ts-ignore`.

rsimon:
Great, thanks!

PR #73, opened 2024-03-05
[DEMO ONLY] Added list of all the block elements to the text annotator example

oleksandr-danylchenko:
# JUST FOR [THE DEMO](https://github.com/recogito/text-annotator-js/issues/74#issue-2169705656), DELETE ME AFTER

### DIFF
--- a/packages/text-annotator/test/index.html
+++ b/packages/text-annotator/test/index.html
@@ -69,6 +69,229 @@
     </div>
 
     <div id="content">
+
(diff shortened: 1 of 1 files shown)


rsimon:
Is this good to close?

oleksandr-danylchenko:
Yeah, I think that it can be closed and it'll still serve the same "just the demo" purpose 👌🏻

PR #75, opened 2024-03-22
Made packages use standalone/peer dependencies

oleksandr-danylchenko:
## Issues
This PR is similar to the https://github.com/annotorious/annotorious/pull/386:
> When I tried using the packages as standalone units I noticed that each sub-package doesn't contain an exhaustive dependencies list that are required for its functioning and building.

---

The dependencies tree for the packages looked like this:
| Package | Dependencies |
|--------|--------|
| `@recogito/text-annotator` | ✅ `@annotorious/core` (was specified as dependency) | 

---

| Package | Dependencies |
|--------|--------|
| `@recogito/text-annotator-react` | ⚠️ `@annotorious/core` (wasn't specified as dependency) | 
| | ⚠️ `@annotorious/react` (wasn't specified as dependency) | 
| | ⚠️ `@recogito/text-annotator`  (wasn't specified as dependency)  | 
| | ⚠️ `@recogito/text-annotator-tei`  (wasn't specified as dependency)  | 

For the `@recogito/text-annotator-react` I moved everything to `dependencies` because those packages are _required_ for the `@recogito/text-annotator-react` to act as a standalone package both on instal in the consuming app or building

---

| Package | Dependencies |
|--------|--------|
| `@recogito/extension-tei` | ⚠️ `@annotorious/core` (wasn't specified as peer or devDependency) | 
| | ⚠️ `@recogito/text-annotator` (wasn't specified as peer or devDependency) |

For the `@recogito/extension-tei` I moved its packages to the `peerDependencies` & `devDependencies` because _it's not a standalone package_. The `extension-tei` can be used only in the project that already has the `text-annotator` initialized

### DIFF
--- a/packages/text-annotator-react/vite.config.js
+++ b/packages/text-annotator-react/vite.config.js
@@ -18,6 +18,7 @@ export default defineConfig(({ command, mode }) => ({
     open: '/test/index.html'
   },
   build: {
+    minify: false,
     lib: {
       entry: './src/index.ts',
       name: 'ReactTextAnnotator',
@@ -26,17 +27,15 @@ export default defineConfig(({ command, mode }) => ({
     },
     rollupOptions: {
       external: [
-        ...Object.keys(packageJson.peerDependencies)
+        ...Object.keys(packageJson.peerDependencies),
+        "@annotorious/core",
+        "@annotorious/react",
+        "@recogito/text-annotator",
+        "@recogito/text-annotator-tei"
       ],
       output: {
         preserveModules: true,
-        assetFileNames: 'react-text-annotator.[ext]',
-        globals: {
-          '@annotorious/react': 'AnnotoriousReact',
-          'openseadragon': 'OpenSeadragon',
-          'react': 'React',
-          'react-dom': 'ReactDOM'
-        }
+        assetFileNames: 'react-text-annotator.[ext]'
       }
     },
     sourcemap: true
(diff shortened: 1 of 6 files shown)


oleksandr-danylchenko:
Ah... Something is still missing: 
image

oleksandr-danylchenko:
I think I might accidentally introduce a bug by extracting packages from the `peerDependecies` to the `dependencies`! As the packages in the prior one were excluded from bundling!

Using the updated `package.json` I found that the `AnnotoriousContext` gets initialized 2 times in the consuming apps❗ 
The 1st time it's created on bundling of the `text-annotator-react`: 
image

And in the app, where I use the ` <Annotorious><TextAnnotator>...</TextAnnotator></Annotorious>`, I see 2 instances of the `AnnotoriousContext` 🤯 
image

That duplicated leads to the hooks' inability to access the proper context with the actual data!

oleksandr-danylchenko:
So I think that the dependencies that have "side-effects" and can be responsible for the creation of the context should be moved to Vite's bundling exclusion list

packages/text-annotator-react/vite.config.js
```diff
@@ -26,19 +27,17 @@ export default defineConfig(({ command, mode }) => ({
...
       ],
       output: {
         preserveModules: true,
-        assetFileNames: 'react-text-annotator.[ext]',
-        globals: {
-          '@annotorious/react': 'AnnotoriousReact',
-          'openseadragon': 'OpenSeadragon',
-          'react': 'React',
-          'react-dom': 'ReactDOM'
-        }
```
oleksandr-danylchenko, line 39: `globals` definition is unused when only the es format is specified

PR #76, opened 2024-03-27
Refactoring + added SPAN-based highlight renderer

rsimon:
## In this PR

This PR:
- makes a few tweaks to the internal store & spatial tree API
- fixes a bug that caused highlight rectangles to get incorrectly indexed, if the page was not scrolled to the top on load (which is usually the case on a page re-load)
- removes redundancy from the rendering code
- adds a new renderer implementation that uses absolutely positioned SPANs, inside a container that is absolutely positioned & aligned with the text content area


packages/text-annotator/src/highlight/canvas/canvasRenderer.ts
```diff
@@ -0,0 +1,121 @@
...
+      offsetRects.forEach(({ x, y, width, height }) => ctx.fillRect(x, y - 2.5, width, height + 5));
+
+      if (style.underlineColor) {
+        ctx.globalAlpha = 1;
+        ctx.strokeStyle = style.underlineColor;
+
+        offsetRects.forEach(({ x, y, width, height }) => {
+          ctx.beginPath();
+          ctx.moveTo(x, y + height + 4);
+          ctx.lineTo(x + width, y + height + 4);
```
oleksandr-danylchenko, line 89: I was about to suggest a change that would allow using the `underlineThickness` & `underlineOffset` for the `CANVAS` renderer:
image

But unfortunately, I'm not sure why the `4` was chosen as a default offset value here 🤔. Should the custom offset override that `4` or should it sum with it?

oleksandr-danylchenko: I _suppose_ that it's _the latter_. Because if I set the offset to `0`, the underline will render over the highlight:
image

rsimon: Hi @oleksandr-danylchenko, I think the Canvas renderer might simply have gotten stuck in mid-development, since I currently switched our client implementation entirely to the SPAN renderer. You are right, of course, that the offset should be configurable, and 0 should align with the box.

I don't have the details in the back of my head exactly, but it's a tricky problem. The annotation highlight size is definitely influenced by various factors, most importantly the CSS `line-height`, and differences between Chrome and FF. However, once the box size is determined and the Canvas renderer uses `ctx.rect()` to draw the box, you should also be able to draw a line that aligns with it. It may not align 100% with the browser selection highlight, but should definitely align with the box drawn on the canvas. Is this not what's happening?

oleksandr-danylchenko: *Added a PR for the `underlineThickness` & `underlineOffset` processing - https://github.com/recogito/text-annotator-js/pull/89

oleksandr-danylchenko: > It may not align 100% with the browser selection highlight

Fortunately, I made the browser's and annotator highlights work separately in my app. The latter will appear only once the selection is finished. So it should be enough to depend only on the `ctx.rect()`.

> However, once the box size is determined and the Canvas renderer uses ctx.rect() to draw the box, you should also be able to draw a line that aligns with it

I'll play with the highlight dimensions a bit to whether the rects sizes can be programmatically synced. So we won't use the "magic" `4`
image

PR #78, opened 2024-03-28
Replaced usage of the `DrawingStyle` with the `HighlightStyleExpression`

oleksandr-danylchenko:
## Issue
In the https://github.com/recogito/text-annotator-js/commit/6b492616d062603bb509350fe1082068b89934e3 the `DrawingStyle | ((annotation: TextAnnotation) => DrawingStyle)` was replaced in majority of places with the text-specific alternative - `HighlightStyleExpression`.

But there were a few places where the `DrawingStyle` still was used. And, the `HighlightStyleExpression` wasn't exported from the `text-annotator` package.

## Changes Made
1. Removed the rest of the `DrawingStyle` usages and replaced it with the `HighlightStyleExpression`
2. Exported highlighting types that should be convenient to import from the package root


rsimon:
Great, thanks!

PR #79, opened 2024-03-28
Decreased `Using renderer` message log level to `debug`

oleksandr-danylchenko:
## Issue
With the `log` level, the info about the currently used renderer appears on every call of the `createTextAnnotator`. This can happen pretty frequently for the React-based apps with `HMR` though. So you start getting a lot of repeating logs:
image


rsimon:
+1. Makes sense. But just to double-confirm: the issue is *just* about hot reloading during dev, right? (Because whenever you see this message, DOM elements get created, annotations get indexed and rendered etc. So if this log shows up during production, something would be wrong with the app.)

oleksandr-danylchenko:
> So if this log shows up during production, something would be wrong with the app.)

I don't that seeing this message multiple times in the console of the prod app _mandatorily_ means that something went wrong 🤔 
For example, in my app, I re-mount the annotator component on the page change using the `key` prop. A fresh annotator and canvas are rendered and seeing the log should be an expected action.

oleksandr-danylchenko:
> But just to double-confirm: the issue is just about hot reloading during dev, right? 

I think that seeing the message too often is _more prominent_ in the dev environment. 

Technically another expected case when you can see that message multiple times on the page is when you manually remount the component with the `key` prop. Which is pretty useful for rendering the canvas from scratch. 

But I suppose that it's not that widespread scenario and seeing the message multiple times can indeed  mean that you apps re-render uncontrollably 👌🏻

PR #83, opened 2024-04-04
Added text-specific `setStyle` override

oleksandr-danylchenko:
## Issue
The `setStyle` method in the `TextAnnotator` was inherited from the `Annotator` type: 
```ts
setStyle(style: DrawingStyleExpression<I> | undefined): void;
```
But the actual type of this method for the text annotator is overridden and looks like:
```ts
setStyle(style: HighlightStyleExpression | undefined): void;
```
due to the following implementation:
https://github.com/recogito/text-annotator-js/blob/263a67ea2096dcc0d16be187393c6b94cecfe9bf/packages/text-annotator/src/TextAnnotator.ts#L85-L86

## Changes Made
Replaced the inherited type def for the `setStyle` with text-specific one

### DIFF
--- a/packages/text-annotator/src/TextAnnotator.ts
+++ b/packages/text-annotator/src/TextAnnotator.ts
@@ -17,6 +17,8 @@ export interface TextAnnotator<E extends unknown = TextAnnotation> extends Annot
 
   element: HTMLElement;
 
+  setStyle(style: HighlightStyleExpression | undefined): void;
+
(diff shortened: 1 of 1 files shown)


oleksandr-danylchenko:
I also looked through the other methods and seems like they match the `Annotator` and don't require an explicit re-defining 👌🏻

rsimon:
Looks good, thanks!

PR #84, opened 2024-04-05
Removed redundant lifecycle generic argument

oleksandr-danylchenko:
## Issue
The `TextAnnotation | E` technically is equal to the `E` and it doesn't change the TS behavior


rsimon:
Makes sense :-) Thanks!

PR #88, opened 2024-04-23
Removed ambiguous `TextAnnotatorPlugin` export

oleksandr-danylchenko:
## Issue
I accidentally added an ambiguous/duplicated definition for the `TextAnnotatorPlugin` within the `@recogito/react-text-annotator` package. Now I would say that users should use [the explicitly re-exported](https://github.com/recogito/text-annotator-js/commit/fca4a185fdec4f1b8f7c6709789fef0f1c4ad6f3) `AnnotoriousPlugin` component


oleksandr-danylchenko:
However, that still might create confusion on why the `TextAnnotator` plugin is used for the plugins definition, but then still gets inserted into the `AnnotoriousPlugin` component 🤔 
I can't think of a good name to draw a line between the `TextAnnotatorPlugin` and "text annotator plugin component". Like it's done for the Annotator/Annotorious

PR #89, opened 2024-04-24
Added `underlineThickness` and `underlineOffset` handling for the `CANVAS` renderer

oleksandr-danylchenko:
## Issue
Previously, only the `underlineColor` could be processed by the `CANVAS` renderer. But not the `underlineThickness` and `underlineOffset`.

### DIFF
--- a/packages/text-annotator/test/index.html
+++ b/packages/text-annotator/test/index.html
@@ -108,16 +108,16 @@
         other gods:
       </p>
 
-<div class="not-annotatable">
-  <h3>Not annotatable block!</h3>
-  <p>
-    "See now, how men lay blame upon us gods for what is after all nothing but their own folly. Look at Aegisthus;
-    he must needs make love to Agamemnon's wife unrighteously and then kill Agamemnon, though he knew it would be
-    the <span class="not-annotatable">death of him; for I sent Mercury to warn him not</span> to do either of these things, inasmuch as Orestes would be
-    sure to take his revenge when he grew up and wanted to return home. Mercury told him this in all good will but
-    he would not listen, and now he has paid for everything in full."
-  </p>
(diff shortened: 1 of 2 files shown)


packages/text-annotator/test/index.html
```diff
@@ -287,12 +287,13 @@ <h3>Not annotatable block!</h3>
...
-        const style = ((annotation, state, z) => ({
-          fillOpacity: 0.2,
-          underlineColor: '#00ff00',
-          underlineOffset: z * 3.5,
-          underlineThickness: 2
-        }));
+        const style = ((annotation, state, z) =>
+	        ({
+		        fillOpacity: 0.2,
+		        underlineColor: '#9b3c06',
```
oleksandr-danylchenko, line 293: Made the color a bit less acidic: 
image

packages/text-annotator/test/index.html
```diff
@@ -287,12 +287,13 @@ <h3>Not annotatable block!</h3>
...
+		        underlineOffset: (z * 3.5) || 0,
```
oleksandr-danylchenko, line 294: Added handling for the case when `z` is not provided, which is relevant for the `CANVAS`

rsimon: Ah - I see you are using our new `z` parameter ;-) It's not really included in the TS function arg definition yet. But we'll keep it as a feature. Happy to see you use it, too :-)

oleksandr-danylchenko: I didn't...

(it was already in the test file 😅. But it was breaking the rendering of the underline when I changed the renderer from the `SPAN` to `CANVAS`. As latter doesn't pass the `z` arg to the styles function)

PR #90, opened 2024-04-24
Expanded `@recogito/react-text-annotator` re-exports

oleksandr-danylchenko:
## Issues
1. The `AnnotoriousPlugin` _component_ was re-exported as the _type_
2. The `TextAnnotation` types from the `@recogito/text-annotator` were not re-exported

### DIFF
--- a/packages/text-annotator-react/src/index.ts
+++ b/packages/text-annotator-react/src/index.ts
@@ -43,6 +43,12 @@ export {
 export type { 
   TextAnnotation,
   TextAnnotationTarget,
+  TextSelector,
+  W3CTextAnnotation,
+  W3CTextAnnotationTarget,
+  W3CTextSelector,
+  W3CAnnotationStylesheet,
+  HighlightStyle,
   TextAnnotator as RecogitoTextAnnotator,
-  TextSelector
+  TextAnnotationStore
 } from '@recogito/text-annotator';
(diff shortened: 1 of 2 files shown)


rsimon:
> Also...
Is there some specific reason to use the following re-export pattern and not re-exporting those props directly?

Do you mean this pattern?

```
import { 
  Origin as _Origin,
  PointerSelectAction as _PointerSelectAction
} from '@annotorious/core'; 

export const Origin = _Origin;
export const PointerSelectAction = _PointerSelectAction;
```

instead of doing this

```
export { 
  Origin,
  PointerSelectAction
} from '@annotorious/core'; 
```

I didn't get the latter pattern to work. The re-exports never got exposed. The explicit import seemed to fix it. However, I believe the reason might actually have been that, previously, `@annotorious/core, `@annotorious/annotorious` etc. weren't listed as `dependencies`. Now that this has changed, the more compact pattern will probably work fine, too.

oleksandr-danylchenko:
Yep, I meant that pattern of re-defining the variables locally. You properly noticed, that the re-export might work out of the box having the `dependencies` included in the package. I'll try it out and commit the re-exports simplification if it works

rsimon:
Wonderful, thanks!

oleksandr-danylchenko:
Seems like it works fine and the methods are properly exposed

Issue #91, opened 2024-05-02
`form`'s `onSubmit` callback doesn't work when the `TextAnnotator` parent is present


oleksandr-danylchenko:
## Issue
I noticed that if my component contains even a basic `form` and it's a descendant of React's `TextAnnotator` - the `onSubmit` callback on it doesn't work 👁️ 

For example, I placed the following form in the example `App.tsx` file:
```tsx
<form
  onSubmit={(e) => {
    e.preventDefault();
    console.log('SUBMIT');
  }}>
  <label>
    Name:
    <input type="text" name="name"/>
  </label>
  <input type="submit" value="Submit"/>
</form>
```
But clicking the submit button has no effect and I don't receive any logs 🤷🏻‍♂️ 

https://github.com/recogito/text-annotator-js/assets/68850090/41ddec5b-eac9-4c51-bbac-63327cb34129

However, when I remove the wrapper - the `form` starts to work as expected again...

https://github.com/recogito/text-annotator-js/assets/68850090/a807aee5-7542-4f9f-99d3-dc7375b3338f

oleksandr-danylchenko:
I also tried commenting out all the listeners registering in the `SelectionHandler`, but it didn't help 🤔

oleksandr-danylchenko:
I suppose to prevent undesired `click` handling, it can be disabled when the click happens in the `not-annotatable` area
https://github.com/recogito/text-annotator-js/blob/3f22130d7342897eed3bc7c93900bdd0415730bc/packages/text-annotator/src/utils/splitAnnotatableRanges.ts#L1

PR #92, opened 2024-05-02
#91 Allowed `click` handling for the not annotatable areas

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/91#issuecomment-2090169107


packages/text-annotator-react/src/TextAnnotator.tsx
oleksandr-danylchenko, line 1: ...just ide spacing...

packages/text-annotator-react/test/App.tsx
oleksandr-danylchenko, line 1: ...just ide spacing...

rsimon:
That makes sense - thanks!

PR #97, opened 2024-05-15
Dev container for node version switching and uniformity

JorenSix:
Hi,

In this pull request you can find a `.devcontainer` folder and minimal configuration to make `text-annotator-js` run.

Dev containers allow developers to quickly setup a sane dev environment with the expected node versions and associated dependencies a dev container helps. See the documentation here https://containers.dev/  

With VS code or other IDEs which support this construct it is practical to do development in such container. It makes it also easy to switch node versions to test older node versions.  A uniform dev environment should also make bugs more easy to replicate or report.

Added to the pull request are two scripts, one is executed once when the container is build, the other each time the container is started. 

It does add a dependency on docker and an additional folder. Thanks for considering this.

Joren

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -1,13 +1,8 @@
 #!/bin/bash
 
+# xdg-open is used to open the browser in the vite start script
+# This is not installed by default in the node image
 sudo apt-get update 
 sudo apt-get install -y xdg-utils
 
-cd packages/extension-tei
-npm install
-
-cd packages/text-annotator
-npm install
-
-cd packages/text-annotator-react
 npm install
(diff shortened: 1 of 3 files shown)


rsimon:
Never heard of it :-) But sounds good in principle. Just one comment: there's no need to `cd` into each individual package folder. (In fact it's bad practice, I believe.) You should run `npm install` only once, in the root folder. npm workspaces will handle the rest.

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,13 @@
...
+sudo apt-get install -y xdg-utils
+
+cd packages/extension-tei
+npm install
+
+cd packages/text-annotator
+npm install
+
+cd packages/text-annotator-react
+npm install
```
oleksandr-danylchenko, line 13:
```suggestion
npm install
```

JorenSix:
Now the npm install thing is only called once in the root folder.

PR #99, opened 2024-05-16
Fixed reading of the `anchorNode: null` 

oleksandr-danylchenko:
## Issue
One of the devs in my team reported that when using the latest version of the `@recogito/text-annotator`, the app started crashing with the following error:
image
He reported that the app  emits a synthetic `selectionchange` event right upon loading with the following shape: 
image
Unfortunately, the `SelectionHandler` `selectionchange` event handler treats the `sel.anchorNode` as it's always present.

## Changes Made
Added `null` value handling for the `sel.anchorNode`

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -52,7 +52,7 @@ export const SelectionHandler = (
     // This is to handle cases where the selection is "hijacked" by another element
     // in a not-annotatable area. A rare case in theory. But rich text editors
     // will like Quill do it...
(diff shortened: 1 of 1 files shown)


rsimon:
Ok, weird. Just for my understanding of the cause. Is this synthetic event coming from your app? Or is this something that's somehow caused by the Text Annotator?

oleksandr-danylchenko:
>  Is this synthetic event coming from your app?

That's something coming from within the app. It fires even w/o Text Annotator initialized on the page

rsimon:
Ok, cool. Just wanted to make sure it's nothing weird & undetected from the Text Annotator. That null check is sensible in any case. Thanks!

PR #100, opened 2024-05-20
Added missing `openseadragon` transitive peer dependency

oleksandr-danylchenko:
## Issue
The `@recogito/react-text-annotator` package depends on the `@annotorious/react` that has the `openseadragon` [defined as a peer](https://github.com/annotorious/annotorious/blob/0b8de66a9723cfa96335e552796dbedee85e7870/packages/annotorious-react/package.json#L35-L44). The `react-text-annotator` doesn't satisfy that dependency, therefore, it should propagate it further as a peer.

### DIFF
--- a/packages/text-annotator-react/package.json
+++ b/packages/text-annotator-react/package.json
@@ -35,8 +35,14 @@
     "vite-tsconfig-paths": "^4.3.2"
   },
   "peerDependencies": {
+    "openseadragon": "^3.0.0 || ^4.0.0",
     "react": "16.8.0 || >=17.x || >=18.x",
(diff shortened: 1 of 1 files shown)


oleksandr-danylchenko:
On the other hand... As it's marked `optional` in the `A9S/react`, then it should be okay to just omit that dependency entirely on the consumer level 🤔

oleksandr-danylchenko:
Although, on the consuming side users will have the following error: 
image
Which makes the dependency not that optional 🤔

rsimon:
Ah, I wasn't aware of optionalDependencies. Looks like this is what we need then instead.

PR #101, opened 2024-05-21
Added `user` to the `TextAnnotatorOptions` list

oleksandr-danylchenko:
## Issue
There was no convenient way to specify the `user` at the text annotator instance creation. Instead, you'd need to call the `setUser` manually.

## Changes Made
1. Added `user` prop to the `TextAnnotatorOptions` type
2. Moved default `user` object resolution to the `fillDefaults` method
3. Added `user` prop reaction to the `TextAnnotator` React wrapper


oleksandr-danylchenko:
Merge candidate? Seems like a pretty small, but beneficial, change 🤔

PR #102, opened 2024-05-21
Added `updated` change upon `currentTarget` update

oleksandr-danylchenko:
## Issue
Previously, the `updated` prop for the `annotation.target` was always `undefined` within the `TextAnnotator`. Although the target changes during the selection.

## Changes Made
Following the [`onChangeSelected` from `SVGDrawingLayer`](https://github.com/annotorious/annotorious/blob/0b8de66a9723cfa96335e552796dbedee85e7870/packages/annotorious-openseadragon/src/annotation/svg/drawing/SVGDrawingLayer.svelte#L137), I added `new Date()` for the `updated` prop on the selection change

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -76,7 +76,8 @@ export const SelectionHandler = (
 
     currentTarget = {
       ...currentTarget,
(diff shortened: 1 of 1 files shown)


rsimon:
Well spotted. Thanks!

Issue #109, opened 2024-06-26
Browser's selection highlight dimensions don't match the emulated highlights


oleksandr-danylchenko:
## Issue
When users select the text, both the default browser's selection highlight and the customly drawn highlight appear. Unfortunately, their dimensions don't match and that leads to the following views across browsers:
| Chrome | Safari | Firefox |
|--------|--------|--------|
| ![image](https://github.com/recogito/text-annotator-js/assets/68850090/6b624c46-f36f-4129-b1c4-9ecfbd36761e) | ![image](https://github.com/recogito/text-annotator-js/assets/68850090/cf6dbcde-0bef-4fcd-8e12-e66fdf3cceed) | ![image](https://github.com/recogito/text-annotator-js/assets/68850090/bed2a5ee-3d53-4234-b175-2b39b275e3f4) | 

That hanging bluish default selection background looks a bit out of place. Ideally, the emulated highlight should match its dimensions. Fortunately, that can be achieved by applying various sizes of compensations based on the browser:
| Chrome | Safari | Firefox |
|--------|--------|--------|
| ![image](https://github.com/recogito/text-annotator-js/assets/68850090/3a09be07-f081-4e0a-a3c3-2a66d86ce4b3) | ![image](https://github.com/recogito/text-annotator-js/assets/68850090/78721c7e-7f21-4a1b-9ffa-aed7290490c6) | ![image](https://github.com/recogito/text-annotator-js/assets/68850090/10d5e742-c677-48c8-a59a-b9a91ef952bc) | 

The padding can be applied in the concrete renderers, as I did for the `canvasRenderer`: 
```ts
// // Chrome
// const selectionHighlightTopCompensation = 3;
// const selectionHighlightBottomCompensation = 7;

// // Safari
// const selectionHighlightTopCompensation = 7;
// const selectionHighlightBottomCompensation = 7;

// Firefox
const selectionHighlightTopCompensation = 0;
const selectionHighlightBottomCompensation = 0;
offsetRects.forEach(({ x, y, width, height }) =>
  ctx.fillRect(
    x,
    y - selectionHighlightTopCompensation ,
    width,
    height + selectionHighlightBottomCompensation
  )
);
```

oleksandr-danylchenko:
I'm surprised that `Firefox` is the only browser that doesn't apply some magical padding to its selection highlight

rsimon:
[...]

PR #110, opened 2024-06-30
#109 Removed selection highlight height compensation in `CANVAS`

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/109#issuecomment-2198591493

## Changes Made
I removed the `selectionHighlightCompensation` variable from the `CANVAS` highlights renderer. It makes the highlight always follow the dimensions reported by the `range.getClientRects`. 

## Demo
| Chrome | Safari | Firefox |
|:----------:|:----------:|:----------:|
 | The default highlight is taller than the rendered one, both at the top & bottom, when applied to the `p` tag. But aligned when applied to `h` | On `p` tag is taller only when applied to 2nd+ row. Aligned on `h` tags and the 1st row of `p` | _Always aligned_ |
 | image image image | image image image | image image image |

### DIFF
--- a/packages/text-annotator/src/highlight/highlights/highlightsRenderer.ts
+++ b/packages/text-annotator/src/highlight/highlights/highlightsRenderer.ts
@@ -21,7 +21,7 @@ const toCSS = (s?: HighlightStyle) => {
   ].filter(Boolean);
 
   return rules.join(';');
-}
+};
 
 export const createRenderer = (): RendererImplementation => {
   const elem = document.createElement('style');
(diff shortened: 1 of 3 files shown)


packages/text-annotator/src/highlight/highlights/highlightsRenderer.ts
oleksandr-danylchenko, line 1: Ignore this file, just auto-formatting

packages/text-annotator/src/highlight/span/spansRenderer.ts
oleksandr-danylchenko, line 1: Ignore this file, just auto-formatting

Issue #112, opened 2024-07-01
`transform: translate(0, 0)` breaks positioning of the highlights using `CANVAS`


oleksandr-danylchenko:
## Issue
When the `transform` property is applied to the annotatable `#content` container - the highlights positioning starts to act weirdly:
- On scroll, the highlights have amplified movement
- On widescreen, the highlights don't render over the text

https://github.com/recogito/text-annotator-js/assets/68850090/223d05a4-f988-41db-a1e1-edd635ce079d

https://github.com/recogito/text-annotator-js/assets/68850090/a817f77a-0844-4fcd-87e5-4fbe64401468

oleksandr-danylchenko:
Seems like the following position compensation lines are problematic when the `transform` is applied:
https://github.com/recogito/text-annotator-js/blob/36958ad904467777e082a4085f17e7415986c569/packages/text-annotator/src/highlight/canvas/canvasRenderer.ts#L55
https://github.com/recogito/text-annotator-js/blob/36958ad904467777e082a4085f17e7415986c569/packages/text-annotator/src/highlight/canvas/canvasRenderer.ts#L83-L88

https://github.com/recogito/text-annotator-js/assets/68850090/fa69b4b4-3f6b-428d-bee0-f33c19f02158

oleksandr-danylchenko:
Aha, so the `transform: translate()` application converts the `#content` into a relative container and makes all the positions calculated against it, not the viewport.

oleksandr-danylchenko:
[Gosh... That's exactly the case I faced...](https://dev.to/salilnaik/the-uncanny-relationship-between-position-fixed-and-transform-property-32f6) 
![image](https://github.com/recogito/text-annotator-js/assets/68850090/6508ea9a-dcc9-42fc-bc8b-154993f9acfe)

oleksandr-danylchenko:
I also tried changing the `position: fixed` and making the `canvas` dimensions match the `container`. But it slaughters the selection/scrolling performance, as it requires the browser to rerender a ton of empty pixels

PR #113, opened 2024-07-01
#112 Fixed mispositioned `canvas` when `container` has `transform` applied

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/112#issuecomment-2199897474

> An element with `position:fixed` is positioned relative to the document (the viewport) which acts as its containing block.
> 
> But, it will NOT always be relative to the document. When any element has `transform`, `filter` or `perspective` property, it acts as a containing block for all its descendants, including the elements whose position is set to fixed.

Therefore, if the `container` has any `transform` applied, even `transform: translate(0, 0);` it'll lead to a mispositioned `canvas` element. It'll be placed not against the viewport, but against the `container` element:
image

https://github.com/recogito/text-annotator-js/assets/68850090/3c3a84f1-34ed-455d-bbcd-2884a2927037

## Changes Made 
Placed the `canvas` element under the `document.body` to ensure that it'll be placed against the viewport


oleksandr-danylchenko:
> We don't use the Canvas renderer

After a bit more testing within my app, I'm also considering the switch to the `SPANS` renderer. As it's much more performant and does have the wobbling effect on scroll:

https://github.com/recogito/text-annotator-js/assets/68850090/aeb75e29-0596-4cae-acf7-3fbdcd26d695

PR #114, opened 2024-07-01
Fixed nearly invisible text under saturated highlights

oleksandr-danylchenko:
### `CANVAS`
| `opacity` | Before | After |
|:--------:|:--------:|:--------:|
| 0.2 | image | image |
| 0.5 | image | image |
| 0.75 | image | image |

### `SPANS`
Not problematic 👌🏻 

### `CSS_HIGHLIGHTS`
Not problematic 👌🏻

### DIFF
--- a/packages/text-annotator/src/TextAnnotator.css
+++ b/packages/text-annotator/src/TextAnnotator.css
@@ -15,15 +15,6 @@ html, body {
   cursor: pointer;
 }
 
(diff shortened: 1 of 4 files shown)


rsimon:
I think this makes sense to add as a default, thanks! Just to add that 'multiply' won't work on dark backgrounds. In this case, the host app will still need to override externally. (`overlay` would normally work well.)

Issue #115, opened 2024-07-01
[`SPANS`] `underlineOffset` is incompatible with the `box-sizing: border-box`


oleksandr-danylchenko:
## Issue
The offset between the highlight and the underline is achieved using the `paddingBottom` of the `span`:
https://github.com/recogito/text-annotator-js/blob/36958ad904467777e082a4085f17e7415986c569/packages/text-annotator/src/highlight/span/spansRenderer.ts#L89-L90
However, that `span` also has the `height` priorly assigned:
 https://github.com/recogito/text-annotator-js/blob/36958ad904467777e082a4085f17e7415986c569/packages/text-annotator/src/highlight/span/spansRenderer.ts#L72

Therefore, when the popular `box-sizing: border-box` style is used, the `paddingBottom` applied after `height` becomes ineffective!

| `box-sizing: content-box` | `box-sizing: border-box` |
|--------|--------|
| ![image](https://github.com/recogito/text-annotator-js/assets/68850090/ad2e522e-5072-429c-8560-f22565f6e898) | ![image](https://github.com/recogito/text-annotator-js/assets/68850090/a6dfaaf9-4ba3-498a-ba62-cc104c4ab18f) |

rsimon:
I think this one should be good to close. I merged the associated pull request.

PR #116, opened 2024-07-01
#115 [`SPANS`] Made `underlineOffset` compatible with the `box-sizing: border-box`

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/115

## Changes Made
1. ~Replaced offset definition using the `paddingBottom` with the `height`. In that way, the height of the highlight span will get properly extended both using the `content-box` and `border-box`.~, see https://github.com/recogito/text-annotator-js/pull/116#issuecomment-2213215373
2. Added `box-sizing: border-box` reset to the `test` examples

## Disclaimer
Most of the changes come from the IDE's formatted that applied 2 space indentation to the `test` files. To see only meaningful differences - hide the whitespace diff:


rsimon:
I might be missing something here. But isn't that just a different way to achieve the same result? Either you apply height + padding, or use `box-sizing: border-box` and change the height. If we want to protect against an external stylesheet unintentionally setting `border-box` on the highlight layer element, we should probably just give it `box-sizing: content-box` explicitely?

oleksandr-danylchenko:
> I might be missing something here. But isn't that just a different way to achieve the same result?

Yeah, it is. The `box-sizing: border-box` is just a more widespread and intuitive sizing approach.

> on the highlight layer element, we should probably just give it box-sizing: content-box explicitly?

However, I think that this suggestion makes sense, and applying `box-sizing: content-box` to the `.r6o-annotation` will our needs

rsimon:
I think this can be closed in favor of https://github.com/recogito/text-annotator-js/commit/dbdbf86b3d09e8db17ce7e21ceed6477e953078f?

oleksandr-danylchenko:
I think this PR can still be beneficial for:
- Traceability of the changes. It explicitly describes that we intend to support the `content-box`
- Makes the `border-box` a default for the test pages, which is relevant for the prevailing majority of the consumers
- Applied the IDE's autoformatting and frees me from constantly tweaking it 😅

Issue #111, opened 2024-06-30
`Ctrl + A` doesn't make the annotation popup to appear


oleksandr-danylchenko:
## Issue
Currently, the `TextAnnotatorPopup` component appears only on the `pointerup` event: 
https://github.com/recogito/text-annotator-js/blob/36958ad904467777e082a4085f17e7415986c569/packages/text-annotator-react/src/TextAnnotatorPopup.tsx#L46-L52
But it ignores the keyboard selection that can be triggered by the `Ctrl + A` shortcut or a11y tools.

I suppose it's a shortcoming of the missing "selection completed" state in the store. Otherwise, we could make the `SelectionHandler` populate it upon meeting some predefined criteria. They can either be `pointerup` (sighted mouse users) or `keyup` (shortcuts and a11y tools support) event

## Demo

https://github.com/recogito/text-annotator-js/assets/68850090/18f0941b-e0f1-4249-b4b6-edebc7008c85

The popup renders only upon clicking somewhere on the page, initiating the `pointerup` event

rsimon:
Addressed in PR #151

PR #118, opened 2024-07-12
Added keyboard text selection support

oleksandr-danylchenko:
## Issue
Previously, only the pointer-based selection was captured and processed by the `SelectionHandler` and the related `TextAnnotatorPopup`. When the user initiated the selection with the `Shift + Arrow` or `Ctrl + A` - it wasn't represented in the annotation selection state. Therefore, we _couldn't know when or whether the annotation popup should appear_.
Instead, for better a11y support, the selection should be able to start and conclude using exclusively keyboard events. The ability to start the selection anywhere on the page is already provided by the [caret browsing](https://www.makeuseof.com/what-is-caret-browsing/) mode.

## Recommendations
Per a11y expert on our team, we received the following recommendations for the selection behavior:
> Ideally, you could release shift, then press it again, and select some more text. This makes tracking Shift (rather than something admittedly awkward like Control Alt/Option h) tricky. But if you could leave focus on the page, and just open the popup when mouse is released or when the keyboard selection starts, and then leave it open through subsequent shift presses and shift clicks, and let users tab to the popup or click a close button or press escape to close the popup, that would work

Also, there were some recommendations on when the popup should be closed:
> Since focus doesn't move into it automatically, I don't think it should be modal. It would be good if it only closes if:
> - the text is unselected
> - the user moves into it and then presses Escape
> - the user moves into and presses a button that would normally close it (not sure it there is a close button - there doesn't have to be one, but it might be helpful)

## Relations
- **This PR depends on the:**
  -  https://github.com/annotorious/annotorious/pull/420, which generalizes the user's selection event capturing. 
  - https://github.com/recogito/text-annotator-js/pull/140, which fixed leaving zombie annotations upon changing the selection direction.
  - https://github.com/recogito/text-annotator-js/pull/149, which fixed reading obsolete `isCollapsed` value when processing `pointerup` event.
- After consulting with a11y UX expert there made a few popup behavior decisions - https://github.com/annotorious/annotorious/issues/419#issuecomment-2225716121
- Addresses the https://github.com/recogito/text-annotator-js/issues/111

## Changes Made
1. Generalized the `lastPointerDown` to just the `lastDownEvent` event, which preserves both the `pointerdown` and `keydown` events. Both can be used to create the annotation selection state within the `onSelectionChange` callback.
2. Added pre-defined focusing behavior into the `TextAnnotatorPopup` component using the [`FloatingFocusManager`](https://arc.net/l/quote/ncllnwjq). Per @rsimon [request](https://github.com/recogito/text-annotator-js/pull/118#issuecomment-2296398240), the automatic focus shift was tuned separately for the mouse and keyboard selection.
For the mouse - the focus automatically goes to the first tabbable element within the floating to reduce the number of clicks required to jot down a note quickly.
For the keyboard - the focus gets shifted to the floating only when a user explicitly presses the `Tab` key. That allows us to make pauses during the selection and not lose the selection range before moving the popup.
3. Replaced `PointerSelectAction` with `UserSelectAction` in re-exports and throughout the code.

### Mouse selection

https://github.com/user-attachments/assets/a437fa52-3132-419b-b419-6e34fd9d7785

### Keyboard selection
1. Selection followed by free arrow key navigation 

https://github.com/user-attachments/assets/c2f43c41-9178-4b11-a08d-e9f529d45f3a

2. Selection followed by `Esc` keypress 

https://github.com/user-attachments/assets/6f441de3-a521-4dda-a072-e3cf396d2faa

3. Selection followed by navigation to the popup and pressing the "Close" button 

https://github.com/user-attachments/assets/aeea3998-2bc4-401d-823e-acb61e1c5667

4. Selection followed by tabbing _**through**_ the popup and bumping into the hint about the dismissal

https://github.com/user-attachments/assets/72f65e33-a0d0-4bad-83cc-17208a23eefa


packages/text-annotator/src/utils/cloneEvents.ts
```diff
@@ -0,0 +1,55 @@
...
+ * 1. It preserves the `target` and `currentTarget` properties.
+ *    Otherwise, they will be `null` when the event is read beyond the handler.
+ * 2. Spread operator can copy only own enumerable properties, not inherited ones.
+ *    Therefore, we need to manually copy the props we're interested in.
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget
+ * @see https://github.com/recogito/text-annotator-js/commit/65d13f3108c429311cf8c2523f6babbbc946013d#r144041390
+ */
+
+export const clonePointerEvent = (event: PointerEvent): PointerEvent => ({
+  ...event,
```
oleksandr-danylchenko, line 12: Even though spreading `event` returns only the `{ isTrusted }` prop - it tricks the Typescript into thinking that the "cloned event" has the `PointerEvent` type

packages/text-annotator/package.json
```diff
@@ -40,7 +40,9 @@
     "@annotorious/core": "^3.0.0-rc.30",
     "colord": "^2.9.3",
     "dequal": "^2.0.3",
+    "hotkeys-js": "^3.13.7",
     "rbush": "^4.0.0",
+    "ts-key-enum": "^2.0.12",
     "uuid": "^10.0.0"
```
rsimon, line 46: I need to review in detail later. But one Q in advance: did you check what the impact is on these dependencies on the overall bundle size? I feel that with basic things, there's always a bit of cost/benefit challenge.

oleksandr-danylchenko: Fortunately, the `ts-key-enum` is implemented as the `const enum` which makes the `Key.Smth` calls replaced w/ real value during the compilation. Therefore, the cost of using the `ts-key-enum` is almost negligible ([source](https://www.npmjs.com/package/ts-key-enum#v3))

> This allows the enum's definition to live entirely in the definition file `https://gitlab.com/nfriend/ts-key-enum/blob/master/Key.enum.d.ts`. Consumers can use this enum without including all of the enum's values in their own JavaScript bundle; only the values referenced in their code will be injected into their bundle.

oleksandr-danylchenko: > `hotkey-js`

> has a reasonable footprint ([~6kB](https://bundlephobia.com/result?p=hotkeys-js)) (gzipped: 2.8kB), and has no dependencies. It should not interfere with any JavaScript libraries or frameworks

oleksandr-danylchenko:
⏫ I discovered a bug that makes the 1-character annotations to create upon changing the selection direction:

https://github.com/user-attachments/assets/8bbba7cf-0b02-457e-a451-ce19a791081b

Working on the fix/workaround for that 👌🏻

packages/text-annotator-react/src/TextAnnotatorPopup.tsx
```diff
@@ -56,17 +55,17 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
...
   const { getFloatingProps } = useInteractions([dismiss, role]);
 
   const selectedKey = selected.map(a => a.annotation.id).join('-');
 
   useEffect(() => {
-    // Ignore all selection changes except those accompanied by a pointer event.
+    // Ignore all selection changes except those accompanied by a user event.
     if (event) {
-      setOpen(selected.length > 0 && event.type === 'pointerup');
+      setOpen(selected.length > 0 && (event.type === 'pointerup' || event.type === 'keydown'));
```
oleksandr-danylchenko, line 66: We should show the popup immediately once the user starts selection. That's because:
1. Prevents having inconsistent workarounds to handle the `up` event for the system shortcuts like `Ctrl + A`, as it can be captured only on `keydown`.
2. Behavior is consistent with the annotating tool of the popular Slab document-creation site

packages/text-annotator-react/test/App.tsx
```diff
@@ -22,13 +22,6 @@ const TestPopup = (props: TextAnnotatorPopupProps) => {
     anno.cancelSelected();
   };
 
-  useEffect(() => {
-    const { current: inputEl } = inputRef;
-    if (!inputEl) return;
-
-    setTimeout(() => inputEl.focus({ preventScroll: true }));
-  }, []);
-
```
oleksandr-danylchenko, line 31: It's better for accessibility to not shift the focus to the popup content right after/during the selection. Here's the a11y expert comment on that matter:

> Ideally, you could release shift, then press it again, and select some more text. This is true also for Mouse users in order to pass a new SC in WCAG 2.2: https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements This is for people who have trouble dragging. This makes tracking Shift (rather than something admittedly awkward like Control Alt/Option h) tricky. But if you could leave focus on the page, and just open the popup when mouse is released or when shift is lifted, and then leave it open through subsequent shift presses and shift clicks, and let users tab to the popup or click a close button or press escape to close the popup, that would work

packages/text-annotator-react/src/TextAnnotatorPopup.tsx
```diff
@@ -56,17 +55,22 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
...
   const selectedKey = selected.map(a => a.annotation.id).join('-');
 
   useEffect(() => {
-    // Ignore all selection changes except those accompanied by a pointer event.
-    if (event) {
-      setOpen(selected.length > 0 && event.type === 'pointerup');
+    // Ignore all selection changes except those accompanied by a user event.
+    if (selected.length > 0 && event) {
+      setOpen(event.type === 'pointerup' || event.type === 'keydown');
+    }
```
oleksandr-danylchenko, line 67: We should show the popup immediately once the user starts selection. That's because:
1. Prevents having inconsistent workarounds to handle the `up` event for the system shortcuts like `Ctrl + A`, as it can be captured only on `keydown`.
2. Behavior is consistent with the annotating tool of the popular Slab document-creation site

rsimon:
We'd have to build a solution that supports both then. I agree on the accessibility aspect. But we used to have it working that way in an earlier system, and there was pretty much a flood of user complaints that the extra click is slowing people down too much. I think there are simply different use cases: some where people read the document and make the occasional annotation once in a while vs. some where people create massive amount of annotations. (E.g. use cases more similar to named entity annotation.)

oleksandr-danylchenko:
> I agree on the accessibility aspect. But we used to have it working that way in an earlier system, and there was pretty much a flood of user complaints that the extra click is slowing people down too much

Okay, then we might wanna be more flexible here 🤔 
For example, when the selection comes from the mouse - we should apply the autofocus to the input area. That would match people's expectations, plus won't break the existing behavior. 
But on the keyboard selection, the autofocus will be removed. Instead, we can have an `aria-live` element that says smth like: "Press Tab to move to &&&& dialog". In that way, the selection won't get abruptly interrupted by the focus shift

oleksandr-danylchenko:
Hello @rsimon 👋🏻 
I've just pushed changes for the "Keyboard Selection 2.0" that preserve your focusing behavior recommendation. See updated Readme to find the demos and change list 🔍

Issue #119, opened 2024-07-12
Jittering positioning of the `TextAnnotatorPopup` on scroll


oleksandr-danylchenko:
## Issue
When the text selection stops higher up in the content and then you start scrolling up/down, the popup starts to jump over the selection range:

https://github.com/user-attachments/assets/db522f4b-8f10-45a2-863e-9de70135cbd5

Personally, I find that kinda distracting, especially on smaller screen sizes: 

https://github.com/user-attachments/assets/6eacb98d-b269-4967-acab-6db893c1f83f

Also, with `autoPlacement` the popup is constantly displayed over the annotated content

## Suggested Changes
We can use a more temperate positioning like `top` + `inline()` middleware. Additionally, we can limit the `crossAxis` scrolling and let the user scroll from the popup. Which feels a bit more pleasant and intuitive (IMHO). 

## Existing solutions
The visual example I like is the Slab.com: 
![image](https://github.com/user-attachments/assets/4ef4ac28-4597-4289-99f4-61aef8042d05)
![image](https://github.com/user-attachments/assets/bc951c99-4f5d-45a1-bb6a-193447f1fb88)

oleksandr-danylchenko:
Moreover, after I added the keyboard selection support in #118, I found that I could accidentally make the page freeze after pressing `Ctrl+A`. I narrowed it down to the popup positioning unable to find a suitable place and running into an endless loop.

PR #120, opened 2024-07-12
#119 Tempered `TextAnnotatorPopup` positioning

oleksandr-danylchenko:
## Changes Made
- Added stable `top` positioning of the popup. It prevents it from covering the annotated content and it's placed pretty intuitively (IMHO)
- Allowed the popup to get scrolled from to prevent it from dragging across the screen over irrelevant text
- Added `useDismiss` hook usage

## Relations
It prevents the page from accidentally freezing when the user selects the text with the `Ctrl+A` (bug I discovered when working on the #118, [see](https://github.com/recogito/text-annotator-js/issues/119#issuecomment-2225895997)

## Demo

https://github.com/user-attachments/assets/2852147a-330d-4306-822c-9be37f8452b6


rsimon:
I like the change. But what if the top of the text is scrolled off screen, and the user selects? Right now, the popup would still pop up above the text, outside the visible area, right? I think it's important that the popup will *always* appear in the visible area - either above or below the text and, at worst, if the selection exceeds both the top and the bottom of the screen, above the annotation. I guess there are behaviors in floating UI that would achieve that?

oleksandr-danylchenko:
> But what if the top of the text is scrolled off the screen, and the user selects? Right now, the popup would still pop up above the text, outside the visible area, right?

Yep, that's correct. It's possible to scroll from the popup and make it non-visible. I believe it can be improved with the [`flip`](https://floating-ui.com/docs/flip) middleware that will place the popup under the selected text on the scroll. Like it's done in Slab: 

https://github.com/user-attachments/assets/5693862c-1592-4a43-a642-ea7404786120

> I think it's important that the popup will always appear in the visible area

I'm not sure whether the popup should still be visible if the user explicitly scrolled from the selected content. Because it removes the visual connection between them 🤔

rsimon:
The issue we face is this: there are sometimes annotations that extend beyond multiple paragraphs, and are larger than the viewport area. 

If the user clicks an annotation that starts above the screen viewport, and ends below the screen viewport, they won't get a popup at all. (Because it would either be above or below the visible screen area.) That's what we need to avoid in our case.

I don't mind if the user actively scrolls the popup out of view. But clicking an annotation always has to open a popup inside the visible viewport area IMO.

oleksandr-danylchenko:
> clicks an annotation that starts above the screen viewport, and ends below the screen viewport, they won't get a popup at all

Aha... I see your point now. I'll try to experiment with the `shift` middleware then 👌🏻

oleksandr-danylchenko:
I found that the most pragmatic solution would be using just the `shift({ mainAxis: false, crossAxis: true, padding: 10 })`. It doesn't introduce weird jumping when the selection area is large, as the `flip` does. And it fulfills the need of having the popup visible for the large references.

rsimon:
YES! I love it!

PR #123, opened 2024-07-24
Made `TextAnnotatorPopup` resilient to the underlying DOM mutations

oleksandr-danylchenko:
## Issue
To position the popup, we used the `range.getBoundingClientRect/getClientRects` methods, which are performant and correct in cases when the annotatable DOM is stable and static. However, when the DOM elements captured within the range mutate  - it collapses and loses its dimensions. That leads to losing the anchoring position for the popup or any other element that depends on that range.

## Changes Made
As we can't rely on the mutable `range` instance, instead we can use the cached `rects` persisted in the `spatialTree`! They are resilient to the underlying DOM mutations and can be manually recalculated on demand. Also, during the recalculation, the ranges can be automatically revived: 
https://github.com/recogito/text-annotator-js/blob/7628ac2a1ce9d17cb3b05f58222d9a3466bfcf97/packages/text-annotator/src/state/spatialTree.ts#L40-L46

## Real-life example
I faced that challenge while integrating the annotator alongside the Speechstream library. The latter highlights the text word-by-word and announces it to the users, working as an a11y/convenience tool. However, its highlighting is implemented in a pretty disruptive way, where the announced paragraphs are constantly being mutated!
| Step | View |
|--------|--------|
| Before announcement  | image |
| After the announcement starts (The highlighted sentence replaces the paragraph) | image |
| Content is placed back (The annotation `range` reference is lost, it became collapsed) | image |
| Further steps where it goes through the sentences word by word | image image ... |

Because of the mutation, the `range` persisted by the annotation was constantly getting collapsed, making it impossible to position related elements by it:
| Text replacement function | Range collapsing |
|--------|--------|
| image | image |


packages/text-annotator/src/state/spatialTree.ts
```diff
@@ -145,7 +149,7 @@ export const createSpatialTree = <T extends TextAnnotation>(store: Store<T>, con
 
   const getAnnotationRects = (id: string): DOMRect[] => {
     const indexed = index.get(id);
-    if (indexed) {
+    if (indexed[0]) {
```
rsimon, line 152: @oleksandr-danylchenko can you say why you made this change? I can see this might be useful as a guard. But wouldn't it break when querying for an ID that's not in the index?

rsimon:
Finally got a chance to review & test this, because we ran into [a different but connected problem ourselves](https://github.com/performant-software/vico/issues/350). This is great, many thanks!

packages/text-annotator/src/state/spatialTree.ts
oleksandr-danylchenko: Yeah, it'll be breaking when asking for a non-existent id. But I didn't think about it as a straight disadvantage. 🤔 
Because the internal code that calls this method will only provide existing ids. And for the consumers that wanna call it from the outside, we set an implicit contract - you must provide an adequate and existing ID for use to return you a value.

rsimon: Hm, if people _can_ call it from outside, they'll call it with all sorts of things. I'd rather protect against that. Is there a case for an annotation being in the index with no rects? Technically, that's currently possible. But I wonder if that shouldn't be filtered out perhaps?

oleksandr-danylchenko: > Is there a case for an annotation being in the index with no rects?

IIRC, I faced that when we used the "selection presence" check to open the popup. On the 1st call of the `getBoundingClientRect` the annotation was in the index but didn't have rects in the spatial tree. But that was quite a while ago 🤔

rsimon: See https://github.com/recogito/text-annotator-js/commit/efa4b06ed87dc595f8ac31514cd45bc16452cd8b

oleksandr-danylchenko: Oh, right... That looks sensible, thanks 🙌🏻

PR #124, opened 2024-07-25
Fixed dimmed text under saturated highlights while using `SPANS` renderer

oleksandr-danylchenko:
## Issue
I noticed that while using the `SPANS` renderer, the text can look dim when you nest the highlights: 
image

## Changes Made
The nature of the issues is the same as it was for the `CANVAS` renderer https://github.com/recogito/text-annotator-js/pull/114. Therefore, I just applied an appropriate `blend-mode` to the `spans` highlight layer:
image


rsimon:
Thanks!

PR #126, opened 2024-08-01
Improve `userSelectAction` naming consistency

oleksandr-danylchenko:
## Issue
Across the parent [annotorious/annotorious](https://github.com/annotorious/annotorious) project we use the `userSelectAction` property name instead of just the `userAction`.
Moreover, changing that value on the React component didn't have any effect.

### DIFF
--- a/packages/text-annotator-react/src/TextAnnotator.tsx
+++ b/packages/text-annotator-react/src/TextAnnotator.tsx
@@ -40,17 +40,11 @@ export const TextAnnotator = <E extends unknown>(props: TextAnnotatorProps<E>) =
     return () => anno.destroy();
   }, [setAnno]);
 
-  useEffect(() => {
-    if (!anno) return;
+  useEffect(() => anno?.setStyle(props.style), [anno, props.style]);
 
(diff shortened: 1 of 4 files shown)


packages/text-annotator-react/src/TextAnnotator.tsx
```diff
@@ -40,17 +40,13 @@ export const TextAnnotator = <E extends unknown>(props: TextAnnotatorProps<E>) =
...
-  }, [anno, props.style]);
+  useEffect(() => anno?.setStyle(props.style), [anno, props.style]);
 
-  useEffect(() => {
-    if (!anno) return;
+  useEffect(() => anno?.setFilter(props.filter), [anno, props.filter]);
 
-    anno.setFilter(props.filter);
-  }, [anno, props.filter]);
+  useEffect(() => anno?.setUserSelectAction(props.userSelectAction), [anno, props.userSelectAction]);
```
oleksandr-danylchenko, line 49: Replicates the https://github.com/annotorious/annotorious/pull/424

Issue #127, opened 2024-08-01
`annotatingEnabled` is not reactive and cannot be changed upon initial mouting


oleksandr-danylchenko:
## Issue
The `annotatingEnabled` property, used by the `SelectionHandler`, gets populated only once upon creating the text annotator instance.
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/TextAnnotator.ts#L29-L32
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/TextAnnotator.ts#L36-L38
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/TextAnnotator.ts#L73

And, unfortunately, there's no callback exposed to update that value during the lifespan of the annotator. Additionally, passing different values for the `annotatingEnabled` as props in the React component- will do nothing 🤷🏻‍♂️ 

## Suggested Changes
Following the example of ["method to change userSelectAction"](https://github.com/annotorious/annotorious/commit/1daff59d1bf7e07df3a27620f131a3def52e1d60) change, we can add the `setUserSelectAction` and preserve the `currentAnnotatingEnabled` within the `SelectionHandler`

PR #128, opened 2024-08-01
#127 Made `annotatingEnabled` property reactive

oleksandr-danylchenko:
## Changes Made
I made the `annotatingEnabled` a dynamic property that can receive values from the external sources via the `setAnnotatingEnabled` (following the ["
API addition: method to change userSelectAction"](https://github.com/annotorious/annotorious/commit/1daff59d1bf7e07df3a27620f131a3def52e1d60) example)

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -10,10 +10,9 @@ import {
   NOT_ANNOTATABLE_SELECTOR
 } from './utils';
 
-export const SelectionHandler = (
+export const createSelectionHandler = (
   container: HTMLElement,
   state: TextAnnotatorState,
-  annotatingEnabled: boolean,
   offsetReferenceSelector?: string
 ) => {
 
@@ -25,6 +24,10 @@ export const SelectionHandler = (
 
   const setFilter = (filter?: Filter) => currentFilter = filter;
 
+  let currentAnnotatingEnabled = true;
+
+  const setAnnotatingEnabled = (enabled: boolean) => currentAnnotatingEnabled = enabled;
+
   const { store, selection } = state;
 
   let currentTarget: TextAnnotationTarget | undefined;
@@ -34,6 +37,8 @@ export const SelectionHandler = (
   let lastPointerDown: PointerEvent | undefined;
 
   const onSelectStart = (evt: PointerEvent) => {
+    if (!currentAnnotatingEnabled) return;
+
     if (!isLeftClick) return;
 
     // Make sure we don't listen to selection changes that were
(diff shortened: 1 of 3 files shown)


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -25,6 +25,18 @@ export const SelectionHandler = (
...
+
+  const setAnnotatingEnabled = (enabled: boolean) => {
+    currentAnnotatingEnabled = enabled;
+    onSelectionChange.clear();
+
+    if (!enabled) {
+      currentTarget = undefined;
+      lastPointerDown = undefined;
+    }
+  };
```
oleksandr-danylchenko, line 38: We must clean the `onSelectionChange` debounced call to prevent it from running on obsolete values from the closure. Otherwise, the `selectionchange` event will still get processed in the deferred call, even though the `currentAnnotatingEnabled` became `false`.

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -113,10 +128,9 @@ export const SelectionHandler = (
       // select events don't have offsetX/offsetY - reuse last up/down)
       selection.userSelect(currentTarget.annotation, lastPointerDown);
     }
-  })
+  }, 10).bind(undefined);
```
oleksandr-danylchenko, line 131: `.bind(undefined)` is required to prevent the `new Error("Debounced method called with different contexts.` exception. When we supply the denounced function directly to the `resize` event listener, it might be invoked with a different context. And the `debounce` library explicitly prohibits that - https://github.com/sindresorhus/debounce/issues/8

oleksandr-danylchenko: That's the wrong approach... The `.bind()` method clears out all the custom properties and methods assigned to the `debounce` function: 
image
That will lead to the following exceptions: 
image

packages/text-annotator/src/utils/debounce.ts
```diff
@@ -1,8 +1,22 @@
...
-		timeoutId = setTimeout(() => func.apply(this, args), delay);
-	}) as T;
+/**
+ * Wraps the `debounce` function from the `debounce` package
+ * to make it context-agnostic.
+ * Otherwise, we won't be able to use it in multiple event listeners simultaneously,
+ * like `window.onresize` and `ResizeObserver`.
+ * @see https://github.com/sindresorhus/debounce/issues/8#issuecomment-2321341074
+ */
+export const debounce: typeof libDebounce = (function_, wait = 10, options) => {
```
oleksandr-danylchenko, line 10: I've just addressed the issue in the `debounce` library - https://github.com/sindresorhus/debounce/pull/43. So I'm waiting for the approval and merge 🙏🏻

packages/text-annotator/src/SelectionHandler.ts
rsimon: I wonder how crucial that is though. Would `annotatingEnabled` get called during a selection in practice? The normal use cases I would see is:
- Setting it once on page init (because the user is getting a page served in "read-only mode")
- "Locking a page" interactively, by the user clicking a "Lock" button. (In which case there wouldn't be a selection.)

Issue #130, opened 2024-08-09
Add `recalculation` completion listener to the `spatialTree`


oleksandr-danylchenko:
## Issue
In the https://github.com/recogito/text-annotator-js/pull/123 PR, I introduced annotation popup positioning via reading the cached values from the Spatial Tree. 

It works pretty great for positioning the popup or any related elements like notes on the static text. But it starts to become a bit flaky when you introduce dynamic content height updates. 

As the popup is mounted on the virtual element the Floating UI doesn't react to the content dimensions updates, and we need to manually call its `update` method:
```ts
useContentContainerResize({ onResize: () => update() });
```
But there's a catch! _The virtual reference element, reading from the spatial tree, might still return obsolete values right upon the resize_!

## Example

Here's an example of how the positioning works against the spatial tree for the notes in my app: 

https://github.com/user-attachments/assets/f750d569-6d45-4563-84c4-295e02b558ac

## Suggestion
The `spatialTree` module can expose the `.on`/`.off`-like handles that will emit events upon the tree recalculation. Then the consumer app will react to them to reposition the required components against the new tree.

It sounds like a much more robust alternative to something like this:
```ts
useContentContainerResize({ onResize: () => setTimeout(update, XXX) });
```
As the timeout cannot be easily predicted and for the large pages with many notes we might not even fit in the limit.

oleksandr-danylchenko:
I see that the `nanoevents` lib is already used for the `UndoStack` in the A9S. And I think that's a nice fit for the spatial tree in terms of practicality.

PR #131, opened 2024-08-09
#130 Added `recalculation` completion listener to the `spatialTree`

oleksandr-danylchenko:
## Changes Made
Following the [`UndoStack`](https://github.com/annotorious/annotorious/blob/main/packages/annotorious-core/src/state/UndoStack.ts) example from A9S, I added the `nanoevents` emitter to the `spatialTree`. 
Currently, we're interested in just a single event type - `recalculated`. Which fires upon the completion of the `recalculate` function

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3132,6 +3132,7 @@
       "version": "9.0.0",
       "resolved": "https://registry.npmjs.org/nanoevents/-/nanoevents-9.0.0.tgz",
       "integrity": "sha512-X8pU7IOpgKXVLPxYUI55ymXc8XuBE+uypfEyEFBtHkD1EX9KavYTVc+vXZHFyHKzA1TaZoVDqklLdQBBrxIuAw==",
+      "license": "MIT",
       "engines": {
         "node": "^18.0.0 || >=20.0.0"
       }
@@ -4371,6 +4372,7 @@
         "@annotorious/core": "^3.0.0-rc.31",
         "colord": "^2.9.3",
         "dequal": "^2.0.3",
+        "nanoevents": "^9.0.0",
         "rbush": "^4.0.0",
         "uuid": "^10.0.0"
       },
--- a/packages/text-annotator/package.json
+++ b/packages/text-annotator/package.json
@@ -40,6 +40,7 @@
     "@annotorious/core": "^3.0.0-rc.31",
     "colord": "^2.9.3",
     "dequal": "^2.0.3",
+    "nanoevents": "^9.0.0",
     "rbush": "^4.0.0",
     "uuid": "^10.0.0"
   }
--- a/packages/text-annotator/src/state/TextAnnotationStore.ts
+++ b/packages/text-annotator/src/state/TextAnnotationStore.ts
@@ -1,5 +1,9 @@
+import type { Unsubscribe } from 'nanoevents';
 import type { Filter, Origin, Store } from '@annotorious/core';
+
 import type { TextAnnotation } from '../model';
+import type { SpatialTreeEvents } from './spatialTree';
+
 
 export interface TextAnnotationStore extends Omit<Store<TextAnnotation>, 'addAnnotation' | 'bulkAddAnnotation'> {
 
@@ -18,6 +22,8 @@ export interface TextAnnotationStore extends Omit<Store<TextAnnotation>, 'addAnn
   getIntersecting(minX: number, minY: number, maxX: number, maxY: number): AnnotationRects[];
 
   recalculatePositions(): void;
+
+  onRecalculatePositions(callback: SpatialTreeEvents['recalculate']): Unsubscribe;
 
 }
 
--- a/packages/text-annotator/src/state/TextAnnotatorState.ts
+++ b/packages/text-annotator/src/state/TextAnnotatorState.ts
@@ -9,7 +9,7 @@ import {
   Origin,
   createViewportState
 } from '@annotorious/core';
-import { createSpatialTree } from './spatialTree';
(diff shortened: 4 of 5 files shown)


rsimon:
Recalculation is a sync operation, though. Might be cleaner to listen to the actual causes than adding this?

oleksandr-danylchenko:
The `recalculate` is invoked as a reaction to the resize of the window or the content within the `baseRenderer`: https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/highlight/baseRenderer.ts#L134-L144
An initial assumption can be that we can add the same listeners to the consuming app and on their firing the spatial tree will already be recalcuted. 
But that is a wrong assumption because we cannot know that the consumer app listeners _will get triggered after the recalculation_.

Therefore, the listener I added in the PR is aimed to provide a method that will allow reacting to a definite completion of the `recalculate` function execution, triggered by the resize observer.

oleksandr-danylchenko:
> Might be cleaner to listen to the actual causes than adding this?

I have a `Note` component that gets positioned next to the highlight. I use FloatingUI to position it, and its virtual element feature. The reference then is the highlight bounds: 
```ts
refs.setPositionReference({
	getBoundingClientRect: () => {
		const bounds = r.state.store.getAnnotationBounds(annotation.id);
		const denormalizedBounds = bounds
			? denormalizeRectWithOffset(bounds, r.element.getBoundingClientRect())
			: new DOMRect();

		const { left: containerLeft } = notesContainer.getBoundingClientRect();
		const { top: annotationTop, height: annotationHeight } = denormalizedBounds;

		return new DOMRect(containerLeft, annotationTop, 0, annotationHeight);
	}
});
```
Floating UI cannot automatically track the changes on screen when the reference is virtual. So I need to manually call the `update` function when the content resizes. Unfortunately, when I _call it immediately upon the resize, the `r.state.store.getAnnotationBounds` still return an obsolete position._ That's because the spatial tree hasn't finished its own recalculate call triggered by the resize

rsimon:
I'm afraid I don't have the time to look into this more closely right now. But might be similar problem to this change I made recently?

https://github.com/recogito/text-annotator-js/commit/0cd547d74d5a52032de2b5760d02f5015f59a8d8

oleksandr-danylchenko:
> I'm afraid I don't have the time to look into this more closely right now

No problem, that's a Friday evening first of all 😅. We have a separate fork to proceed with the development, but I'm glad to keep you in a loop of changes that might be beneficial for the annotator in general.

oleksandr-danylchenko:
> I'm afraid I don't have the time to look into this more closely right now (and I'm only on mobile). But might be similar problem to this change I made recently?
> 
> [0cd547d](https://github.com/recogito/text-annotator-js/commit/0cd547d74d5a52032de2b5760d02f5015f59a8d8)

I had the same code before in my app and works perfectly when the virtual reference element is consulting the native browser's `range`s 👌🏻 

---

However, we couldn't use them anymore, due to the 3rd party integration in our app that wildly mutates the underlying DOM, making the captured ranges collapse. That's why I switched to the spatial tree and its cached boundaries. (https://github.com/recogito/text-annotator-js/pull/123)

rsimon:
Either way: IMO a component shouldn't trigger an async-like event after completing a sync operation. At least it should be the component that _triggers_ the sync op. (Might already be the case in your PR. As I said: can't look into it right now yet :-)

oleksandr-danylchenko:
> IMO a component shouldn't trigger an async-like event after completing a sync operation. At least it should be the component that triggers the sync op

That's fair, I like this suggestion 👍🏻
I'll see how I can tweak the emitter to match it, thanks!

rsimon:
> I had the same code before in my app and it works perfectly when the virtual reference element is consulting the native browser's ranges 👌🏻

Got it,  thanks!

oleksandr-danylchenko:
> IMO a component shouldn't trigger an async-like event after completing a sync operation. At least it should be the component that triggers the sync op

I think that having the emission from the `spatialTree` _fits the usage patterns_ we already have in the A9S `Store` and the `UndoStack`. 

Although those components don't have asynchronous methods, they still emit the events as a reaction to their functions's invocation. [Example](https://github.com/annotorious/annotorious/blob/e6a5ad1bbebc3fd2eeda343ed265490e4047dd48/packages/annotorious-core/src/state/Store.ts#L330-334): 
image

Or [another one](https://github.com/annotorious/annotorious/blob/e6a5ad1bbebc3fd2eeda343ed265490e4047dd48/packages/annotorious-core/src/state/UndoStack.ts#L111-L125):
image

So I would think about the spatial tree as the "store"-like module from which we can receive events about recalculation / update / etc.

---

I also thought about moving the event emission either to the `Store`'s `recalculatePositions`: https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/state/TextAnnotatorState.ts#L131
or next to the `recalculatePositions` call from the `baseRenderer`:
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/highlight/baseRenderer.ts#L134-L142
But I'm not extremely fond of them. Because then we can have a scenario when the `recalculate` gets invoked directly on the spatial tree instance, but none of the `onRecalculatePositions` listeners won't be notified 🤔

Therefore, I would _recommend_ (but not insist yet 😅) leaving the `emitter.emit('recalculate')` within the `spatialTree`.

PR #134, opened 2024-08-13
#132 Fixed unexpected ability to create annotations outside of the annotatable container

### LINKED ISSUE #132: Content is annotatable outside of the annotator `container` ⚠️
## Issue
It's possible to create annotation on the text that's outside of the annotatable content container: 

https://github.com/user-attachments/assets/e1366c9e-9c6d-4105-acb9-0254e4e8e1e1

When the selection starts on annotatable content at least once, all the surrounding content becomes annotatable too! In the test app it's just another text, but in the real app users can annotate headers/footers/information elements, etc.

Instead, it would be expected that the annotation is handled only within the container initialized for the annotator instance: https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/test/index.html#L288
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/test/index.html#L298-L303

oleksandr-danylchenko:
## Issue
Fixes https://github.com/recogito/text-annotator-js/issues/132

## Changes Made
1. Added `trimRangeToContainer` method that trims the range down to a specified element when its edges "stick out".
2. Added trimming of the global selection range to the annotatable container in the selection handler.
Therefore, users can no longer annotate content outside of the container over which the annotator was initialized ✔️ 

## Demo

https://github.com/user-attachments/assets/66ee5681-dcdc-483a-92f1-70961424d0d3

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -7,6 +7,7 @@ import {
   splitAnnotatableRanges,
   rangeToSelector,
   isWhitespaceOrEmpty,
+  trimRangeToContainer,
   NOT_ANNOTATABLE_SELECTOR
 } from './utils';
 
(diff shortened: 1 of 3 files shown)


rsimon:
That's a nice addition - thanks!

Issue #133, opened 2024-08-13
Annotation isn't dismissed clicking outside of the annotator `container` ⚠️


oleksandr-danylchenko:
## Issue
When you have a selection within the annotatable container and click outside - the selected annotation won't be dismissed.
Instead, the `userSelect` method is called on it:

https://github.com/user-attachments/assets/8d9bc14c-f388-486d-92bc-a22e53f6eb5e

That's because the `pointerDown` method is initialized only on the `container`, but the `pointerUp` is initialized on the whole document. Therefore, the `currentTarget` is still present when the latter is executed, and the `evt.timeStamp - lastPointerDown.timeStamp` is always too large:
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/SelectionHandler.ts#L151-L157

rsimon:
I've also been back and forth with this. I think we have different UI requirements here, and I can see valid use cases for both. (Which may mean we might need to make this a config option.)

In our case we have a whole bunch of UI components around the annotatable area. Things like setting annotation filters, changing color coding etc. Most of the time, users would expect that nothing should change about the selection when interacting with those parts of the UI. I.e. in this case, clicks outside the annotatable area should simply be ignored completely. 

I guess we could use a CSS selector like `not-annotatable` to mark such areas. (Or maybe `not-annotatable` is actually enough?) But I wonder if this wouldn't lead to some kind of error prone patchwork where you start sprinkling (and forgetting to sprinkle) these CSS classes all over the place?

oleksandr-danylchenko:
> Or maybe `not-annotatable` is actually enough?)

Applying the `not-annotatable` class _should be enough_. It can be used just for the parent container of the controls, so we'll exclude all of them at once. 

> But I wonder if this wouldn't lead to some error-prone patchwork where you start sprinkling (and forgetting to sprinkle) these CSS classes all over the place?

I also thought about this, but I concluded that we'll always have _a fairly limited set of known controls_ that we can always cover with the `not-annotatable` containers. Therefore, it seemed like a good enough compromise to me.

oleksandr-danylchenko:
[...]

PR #135, opened 2024-08-13
#133 Fixed missing annotation dismissal clicking outside of the annotator `container`

oleksandr-danylchenko:
## Changes Made
To capture the actual `lastPointerDown` event on the page and make the correct `timeDifference` calculations, I added `pointerdown` event to the `document` scope.

In that way, the "click" will be properly recognized in the `pointerup` handler and the selection will be dismissed. As there will be no annotation under the pointer.

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -124,7 +124,7 @@ export const SelectionHandler = (
     isLeftClick = evt.button === 0;
   }
 
-  container.addEventListener('pointerdown', onPointerDown);
+  document.addEventListener('pointerdown', onPointerDown);
 
   const onPointerUp = (evt: PointerEvent) => {
     const annotatable = !(evt.target as Node).parentElement?.closest(NOT_ANNOTATABLE_SELECTOR);
@@ -135,7 +135,11 @@ export const SelectionHandler = (
     const clickSelect = () => {
       const { x, y } = container.getBoundingClientRect();
 
-      const hovered = store.getAt(evt.clientX - x, evt.clientY - y, currentFilter);
+      const hovered =
+        evt.target instanceof Node &&
+        container.contains(evt.target) &&
+        store.getAt(evt.clientX - x, evt.clientY - y, currentFilter);
+
       if (hovered) {
         const { selected } = selection;
 
@@ -163,7 +167,7 @@ export const SelectionHandler = (
     container.removeEventListener('selectstart', onSelectStart);
     document.removeEventListener('selectionchange', onSelectionChange);
     
-    container.removeEventListener('pointerdown', onPointerDown);
+    document.removeEventListener('pointerdown', onPointerDown);
     document.removeEventListener('pointerup', onPointerUp);
   }
 


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -135,7 +135,11 @@ export const SelectionHandler = (
     const clickSelect = () => {
       const { x, y } = container.getBoundingClientRect();
 
-      const hovered = store.getAt(evt.clientX - x, evt.clientY - y, currentFilter);
+      const hovered =
+        evt.target instanceof Node &&
+        container.contains(evt.target) &&
```
oleksandr-danylchenko, line 140: Added the `container.contains` as a small optimization that will immediately dismiss clicks outside of the `container`. So we won't need to execute the relatively heavier `getAt` method

rsimon:
Thanks - another nice & useful one!

rsimon:
I had to revert this change, unfortunately. We have a 'Delete' button for the current annotation (outside the annotatable area) and that was no longer clickable after this change.

(Interestingly, the onClick handler no longer got called at all. Would need to investigate further.)

oleksandr-danylchenko:
> We have a 'Delete' button for the current annotation (outside the annotatable area). It is only shown when the annotation is selected. Because the selection got discarded before the click event, clicking the button had no effect. (The button was unmounted before the click event.)

Right... Until the `click` happens, the selection gets dismissed...
I believe that it might be mitigated by explicitly checking what element is the target of the `pointerdown` event and preventing selection dismissal if a specific classname/attribute is present.
From the consumer perspective, it would act like the `not-annotatable` class, that they should apply manually based on their specific needs. The library will just provide a convenient default behavior that will fit "most" of the cases.

oleksandr-danylchenko:
Unfortunately, moving the `pointerdown` listener directly to the `document` has just bitten me too! If I capture it on an `input` or a `button` and quickly drag to the textual content - the selection will be dismissed! 

https://github.com/user-attachments/assets/9fe2762b-b0d2-4b3b-925e-b11ea8117752

That happens because starting the selection within the interactive element - traps the selection range within it. Therefore, on the `pointerup` event the selection will always be collapsed. And as we capture the `lastPointerDown` even on not-annotatable elements - the `timeDifference` will be pretty small!

oleksandr-danylchenko:
However, I think I have an idea of how to: 
1. resolve #133, dismiss selection when clicking outside of the `container`
2. don't break the "Delete" button functionality you mentioned
3. resolve the unexpected dismissal when the mouse is being dragged from the interactive element

**It's capturing the `lastPointerEvent` only on the elements that are not contained within the `not-annotatable` tree!**
That would be helpful for the aforementioned issues because:
1. When a user clicks somewhere outside the `container` - such a target usually won't be located within the `not-annotatable` parent. Therefore, the `pointerdown` event will be preserved and will create a proper `timeDifference` in the `pointerup` handler to dismiss the selection!
2. If you explicitly apply the `not-annotatable` to the `Delete` button - the `pointerdown` will not get captured on it. So the `timeDifference` in the `pointerup` handler would be sufficient to not treat the `pointerup` as the "click".
3. Essentially the same as 2., but we already have the `not-annotatable` on the desired components like the popup or the side notes (in the consuming app)

Does it sound right to you, @rsimon?

PR #138, opened 2024-08-16
Added default `z-index` for the span highlights

oleksandr-danylchenko:
## Issue
The `SPAN`'s `.r6o-span-highlight-layer` should be brought up above the regular textual content on the page. That will prevent elements with the background from being displayed over the highlights.

## Example
| W/o `z-index` | W/ `z-index: 1` |
|--------|--------|
| image | image | 

## Changes Made
Added `z-index: 1` to put the `.r6o-span-highlight-layer` out of the nesting stack. Similar treatment has already been done to the canvas: 
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/highlight/canvas/canvasRenderer.css#L11-L14

### DIFF
--- a/packages/text-annotator/src/highlight/span/spansRenderer.css
+++ b/packages/text-annotator/src/highlight/span/spansRenderer.css
@@ -1,11 +1,12 @@
 .r6o-annotatable .r6o-span-highlight-layer {
-  height: 100%;
-  left: 0;
-  mix-blend-mode: multiply;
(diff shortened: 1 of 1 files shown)


rsimon:
I'm not sure I understand the exact problem in the example. But it's definitely a sensible fix :-)

PR #140, opened 2024-08-16
Added annotation discarding when the selection collapses

oleksandr-danylchenko:
## Changes Made
I added explicit deletion of the annotation that was created during the selection that collapsed. It removes zombie 1-char annotation when you switch the selection direction.


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -72,17 +72,31 @@ export const SelectionHandler = (
...
+     * The selection range got collapsed during the selecting process.
+     * The previously created annotation isn't relevant anymore and can be discarded
+     *
+     * @see https://github.com/recogito/text-annotator-js/issues/139
+     */
+    if (sel.isCollapsed && store.getAnnotation(currentTarget.annotation)) {
+      selection.clear();
+      store.deleteAnnotation(currentTarget.annotation);
+      return;
+    }
```
oleksandr-danylchenko, line 89: I intentionally don't reset the `currentTarget` here, because during the mouse selection cycle, the `selectstart` won't be dispatched another time to create a new target instance.

oleksandr-danylchenko:
Discovered a bug that falsely deletes annotations in Firefox and Safari, but not Chrome:

https://github.com/user-attachments/assets/1171ce99-e8a7-4cb4-8994-b5fe526e0932

https://github.com/user-attachments/assets/6cb05779-2c28-4d90-ae7d-ad1379045d4e

The only difference is the result of the `store.getAnnotation(currentTarget.annotation)` check. In Chrome - it returns `undefined`, skipping the deletion. But in FF/Safari - it returns an annotation, so the deletion happens.

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -122,6 +132,7 @@ export const SelectionHandler = (
     lastPointerDown = { ...evt, target, timeStamp, offsetX, offsetY, type };
 
     isLeftClick = evt.button === 0;
+    currentTarget = undefined;
```
oleksandr-danylchenko, line 135: Having the `currentTarget` cleaned up on the `pointerdown` event guarantees that on the deletion check, we won't accidentally refer to the dismissed annotation

oleksandr-danylchenko: UPD: It led to a regression while using the touch devices. When user scrolls while having an active selection range, they cannot proceed selecting anymore

oleksandr-danylchenko: Another madness piece... In Firefox, the `selectstart` event gets dispatched AFTER the `selectionstart` 🤯 
image

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -66,12 +64,38 @@ export const SelectionHandler = (
...
-    if (sel.isCollapsed || !isLeftClick || !currentTarget) return;
+    } else if (sel.isCollapsed && timeDifference < CLICK_TIMEOUT) {
+
+      /*
+       Firefox doesn't fire the 'selectstart' when user clicks
+       over the text, which collapses the selection
+      */
+      onSelectStart(lastPointerDown);
+
+    }
```
oleksandr-danylchenko, line 83: Those lines unify Firefox's handling of the click over the textual content with the rest of the browsers. It's _the only browser_ that has that doesn't dispatch the `selectstart` when the selection range collapses.
Therefore, when the user clicks somewhere on the screen upon finishing the selection, that new annotation will get falsely deleted [here](https://github.com/recogito/text-annotator-js/pull/140/files#diff-1e072053802159d91da89e331e1fe166da3cb395a3fc6ab3d6317b0510ea17e0R94-R98), because the `currentTarget` will still refer to it. But in Chrome/Safari, the `selectstart` event will be dispatched, and the `currentTarget` will get updated to the one that doesn't exist in the store yet.
Demos - https://github.com/recogito/text-annotator-js/pull/140#issuecomment-2307515070

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -147,13 +171,14 @@ export const SelectionHandler = (
...
 
     // Just a click, not a selection
-    if (document.getSelection().isCollapsed && timeDifference < 300) {
+    if (sel?.isCollapsed && timeDifference < CLICK_TIMEOUT) {
       currentTarget = undefined;
       clickSelect();
-    } else if (currentTarget) {
+    } else if (currentTarget && store.getAnnotation(currentTarget.annotation)) {
       selection.userSelect(currentTarget.annotation, evt);
     }
```
oleksandr-danylchenko, line 183: Prevents selection of the `currentTarget` that could have already been deleted in the `selectionchange` event handler.

PR #141, opened 2024-08-22
Added accessible announcement for the `TextAnnotatorPopup` appearance

oleksandr-danylchenko:
## Issue
When the `TextAnnotatorPopup` appears during the keyboard selection, the focus isn't immediately shifted to it ([reason](https://github.com/recogito/text-annotator-js/pull/118#issuecomment-2296427645)). Instead, a user should explicitly Tab to enter into it. 
_However, there's no way to know when or whether the popup appeared for the screen reader users._ 

## Suggestion
A navigation hint message should be explicitly announced using the live area. The message should appear after a user idles for at least `700ms` to prevent it from being interrupted by system announcements. The hint should be announced once per selection so as not to be too annoying.

## Changes Made
1. Added `@react-aria/live-announcer` dependency ([source](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/live-announcer/src/LiveAnnouncer.tsx)), that allows to announce arbitrary messages on the page conveniently.
2. Added listeners for the user's idling that triggers the announcement.
3. Made the `message` configurable.

## Relations
This PR depends on the Keyboard Selection Support one - https://github.com/recogito/text-annotator-js/pull/118. Therefore, it includes all of its changes within.


oleksandr-danylchenko:
After #151 the selection behavior now opens the popup only on the selection "pause"/"completion" - we can make the announcer say the navigation message on its opening

oleksandr-danylchenko:
I discovered that enabling this hook on the popup breaks the Mac's "Look up %word%" feature:

https://github.com/user-attachments/assets/28b07bff-85ac-4701-9e0d-b6cca726499a

PR #146, opened 2024-09-02
Fixed popup content props usage

oleksandr-danylchenko:
## Issue
In https://github.com/recogito/text-annotator-js/commit/7fafa46a8c3522753c9cace1027c4f02369d9a32 the interface of the `TextAnnotatorPopup` content props got updated in correspondence to the image popup. 
However, the test files were not updated along and it led to the exceptions when you try running the `text-annotator-react` examples on `main`.

### DIFF
--- a/packages/text-annotator-react/src/TextAnnotatorPopup.tsx
+++ b/packages/text-annotator-react/src/TextAnnotatorPopup.tsx
@@ -19,7 +19,7 @@ interface TextAnnotationPopupProps {
 
 }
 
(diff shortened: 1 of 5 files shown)


oleksandr-danylchenko:
I also noticed that the popup is displayed below the highlights. I suppose that's a regression from the https://github.com/recogito/text-annotator-js/pull/138 that missed somehow 🤔 
image

rsimon:
Thanks!

PR #150, opened 2024-09-27
⚠️ Fixed processing of the selection range made totally outside of the container ⚠️

oleksandr-danylchenko:
## Issue
When the selection range was made entirely outside of the container - we should treat it as a collapsed one and discard it. However, this wasn't the case with the `trimRangeToContainer` method I introduced in https://github.com/recogito/text-annotator-js/pull/134. Instead, it would make the outside range cover the container ⚠️ 

image

###### (Soomo Staged, 22.09.25)


rsimon:
E.g. something like the following the below?

* Restrict the hotkey listener to the container.
* When `onSelectAll` gets triggered, prevent full-page selection, select the container instead - which will trigger an annotation across the full annotatable area (with `not-annotatable` regions removed).

```
hotkeys(SELECT_ALL, { element: container, keydown: true, keyup: false }, evt => {
    lastDownEvent = cloneKeyboardEvent(evt);
    onSelectAll(evt);
});
``` 
and 

```
const onSelectAll = (evt: KeyboardEvent) => {
    evt.preventDefault();

    const range = document.createRange();
    range.selectNodeContents(container);

    const s = window.getSelection();
    s.removeAllRanges();
    s.addRange(range);

    // ... 
    
}
```

oleksandr-danylchenko:
> no longer seem to work if it means that the selection starts in non-annotatable space

I think that there's another UX quirk that appeared due to handling everything outside as not-annotatable. When a user starts their selection from the page's gutter, the annotation won't be preserved due to the same `isNotAnnotatable(container, evt.target as Node)` check:

https://github.com/user-attachments/assets/7f8ec9c0-d26b-4888-9d73-2be2b80baccd

PR #151, opened 2024-10-02
Revised Selection Behavior

rsimon:
## In this PR

This PR revises the selection behavior fundamentally. Instead of creating an annotation immediately, the SelectionHandler now waits until the (native browser text-)selection is 'complete'. Because there is not actually a 'complete' event, there are different approaches, depending on platform and interaction mode:

- For __mouse interaction__ a complete selection is determined by a `pointerup` event.
- For __keyboard interaction__ a complete selection is determined by the `keyup` event of the Shift key. (Note that selection remains editable - unless popup intervenes.)
- `Ctrl` + `A` (or `Cmd` + `A`) is handled separately, as a single operation.
- __iOS/iPad__: complete selection is determined by `pointerup` or `contextmenu` event.
- __Android__: complete selection is determined by the `contextmenu` event. 

__Note:__ in all cases, the selection remains editable after "completion", either by using Shift+Arrow keys (desktop) or by dragging the handlebars (on mobile/touch). 

### React Popup

I also modified the React `TextAnnotatorPopup` with the following behaviors: 

- On mobile, the popup is oriented __below__ the annotation (instead of the default above), since it would otherwise be obstructed by the default system context menu on iOS and Android.
- On mobile, the popup does not receive initial focus. The user must explicitly tap into it. This way, the selection handlebars remain activated, and it is still possible to alter the selection after the popup has opened.

__Warning:__ mobile mode is detected through `navigator.userAgent`. It appears that Firefox currently __does not include sufficient information to determine whether it is running on iOS__. Therefore, FF/iOS currently behaves like on the desktop, which also means it will lose the selection handlebars as soon as the `<TextAnnotatorPopup>` opens.

### DIFF
--- a/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
+++ b/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
@@ -1,4 +1,6 @@
-import React, { PointerEvent, ReactNode, useCallback, useEffect, useState } from 'react';
+import { PointerEvent, ReactNode, useCallback, useEffect, useState } from 'react';
+import { useAnnotator, useSelection } from '@annotorious/react';
+import type { TextAnnotation, TextAnnotator } from '@recogito/text-annotator';
 import {
   autoUpdate,
   flip,
@@ -12,9 +14,6 @@ import {
   useInteractions,
   useRole
 } from '@floating-ui/react';
-
-import { useAnnotator, useSelection } from '@annotorious/react';
-import type { TextAnnotation, TextAnnotator } from '@recogito/text-annotator';
 
 import './TextAnnotatorPopup.css';
 
@@ -162,4 +161,4 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
     </FloatingPortal>
   ) : null;
 
-};
+}
--- a/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
+++ b/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
@@ -1,13 +1,11 @@
-import { PointerEvent, ReactNode, useCallback, useEffect, useState } from 'react';
+import { ReactNode, useCallback, useEffect, useState, PointerEvent } from 'react';
 import { useAnnotator, useSelection } from '@annotorious/react';
-import type { TextAnnotation, TextAnnotator } from '@recogito/text-annotator';
+import { type TextAnnotation, type TextAnnotator } from '@recogito/text-annotator';
 import {
   autoUpdate,
-  flip,
-  FloatingFocusManager,
-  FloatingPortal,
   inline,
   offset,
+  flip,
   shift,
   useDismiss,
   useFloating,
@@ -15,18 +13,16 @@ import {
   useRole
 } from '@floating-ui/react';
 
-import './TextAnnotatorPopup.css';
-
 interface TextAnnotationPopupProps {
 
   popup(props: TextAnnotationPopupContentProps): ReactNode;
 
 }
 
-export interface TextAnnotationPopupContentProps {
+interface TextAnnotationPopupContentProps {
 
   annotation: TextAnnotation;
-
+  
   editable?: boolean;
 
   event?: PointerEvent;
@@ -37,27 +33,21 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
 
   const r = useAnnotator<TextAnnotator>();
 
-  const { selected, event } = useSelection<TextAnnotation>();
+  const { selected } = useSelection<TextAnnotation>();
+
   const annotation = selected[0]?.annotation;
 
   const [isOpen, setOpen] = useState(selected?.length > 0);
 
-  const handleClose = () => {
-    r?.cancelSelected();
-  };
-
   const { refs, floatingStyles, update, context } = useFloating({
     placement: 'top',
     open: isOpen,
-    onOpenChange: (open, _event, reason) => {
+    /* onOpenChange: (open, _event, reason) => {
       setOpen(open);
-
-      if (!open) {
-        if (reason === 'escape-key' || reason === 'focus-out') {
-          r?.cancelSelected();
-        }
+      if (!open && reason === 'escape-key') {
+        r?.cancelSelected();
       }
-    },
+    }, */
     middleware: [
       offset(10),
       inline(),
@@ -68,41 +58,33 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
   });
 
   const dismiss = useDismiss(context);
-  const role = useRole(context, { role: 'dialog' });
+
+  const role = useRole(context, { role: 'tooltip' });
+  
   const { getFloatingProps } = useInteractions([dismiss, role]);
 
   const selectedKey = selected.map(a => a.annotation.id).join('-');
 
   useEffect(() => {
-    // Ignore all selection changes except those accompanied by a user event.
-    if (selected.length > 0) { // && event) {
-      setOpen(true); // event.type === 'pointerup' || event.type === 'keydown');
-    }
-  }, [selectedKey /*, event */]);
+    // Ignore all selection changes except those accompanied by a pointer event.
+    setOpen(selected.length > 0);
+  }, [selectedKey]);
 
   useEffect(() => {
-    // Close the popup if the selection is cleared
-    if (selected.length === 0 && isOpen) {
-      setOpen(false);
-    }
-  }, [isOpen, selectedKey]);
+    if (!isOpen || !annotation) return;
 
-  useEffect(() => {
-    if (isOpen && annotation) {
-      const {
-        target: {
-          selector: [{ range }]
-        }
-      } = annotation;
+    if (!annotation.target.selector || annotation.target.selector.length === 0) return;
 
-      refs.setPositionReference({
-        getBoundingClientRect: range.getBoundingClientRect.bind(range),
-        getClientRects: range.getClientRects.bind(range)
-      });
-    } else {
-      // Don't leave the reference depending on the previously selected annotation
-      refs.setPositionReference(null);
-    }
+    const {
+      target: {
+        selector: [{ range }]
+      }
+    } = annotation;
+
+    refs.setPositionReference({
+      getBoundingClientRect: () => range.getBoundingClientRect(),
+      getClientRects:() => range.getClientRects()
+    });
   }, [isOpen, annotation, refs]);
 
   // Prevent text-annotator from handling the irrelevant events triggered from the popup
@@ -122,43 +104,21 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
     return () => {
       mutationObserver.disconnect();
       window.document.removeEventListener('scroll', update, true);
-    };
+    }
   }, [update]);
 
   return isOpen && selected.length > 0 ? (
-    <FloatingPortal>
-      <FloatingFocusManager
-        context={context}
-        modal={false}
-        closeOnFocusOut={true}
-        initialFocus={
-          /**
-           * Don't shift focus to the floating element
-           * when the selection performed with the keyboard
-           */
-          event?.type === 'keydown' ? -1 : 0
-        }
-        returnFocus={false}
-      >
-        <div
-          className="annotation-popup text-annotation-popup not-annotatable"
-          ref={refs.setFloating}
-          style={floatingStyles}
-          {...getFloatingProps()}
-          {...getStopEventsPropagationProps()}>
-          {props.popup({
-            annotation: selected[0].annotation,
-            editable: selected[0].editable,
-            event
-          })}
-
-          {/* It lets keyboard/sr users to know that the dialog closes when they focus out of it */}
-          <button className="popup-close-message" onClick={handleClose}>
-            This dialog closes when you leave it.
-          </button>
-        </div>
-      </FloatingFocusManager>
-    </FloatingPortal>
+    <div
+      className="annotation-popup text-annotation-popup not-annotatable"
+      ref={refs.setFloating}
+      style={floatingStyles}
+      {...getFloatingProps()}
+      {...getStopEventsPropagationProps()}>
+      {props.popup({
+        annotation: selected[0].annotation,
+        editable: selected[0].editable
+       })}
+    </div>
   ) : null;
 
 }
--- a/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
+++ b/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
@@ -73,12 +73,13 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
   const { getFloatingProps } = useInteractions([dismiss, role]);
 
   const selectedKey = selected.map(a => a.annotation.id).join('-');
+
   useEffect(() => {
     // Ignore all selection changes except those accompanied by a user event.
-    if (selected.length > 0 && event) {
-      setOpen(event.type === 'pointerup' || event.type === 'keydown');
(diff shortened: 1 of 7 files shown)


rsimon:
Hi @oleksandr-danylchenko,

I'm about to merge this into `main` soon. But wanted to give you a heads-up and chance to look over the changes. I think this should work well on all platforms. But do let me know if you have any doubts, or find something that I missed.

I so far tested this on FF + Chrome (Mac, iPad, Android) and will do another round of testing on FF/Chrome/Edge on Linux tomorrow, with a hybrid desktop/touch computer.

oleksandr-danylchenko:
Thank you 🙌🏻🙌🏻🙌🏻
I'll take a look tomorrow

rsimon:
__Update:__ I added a few more tweaks. I could find a perfect solution for CTRL/CMD-A, unfortunately. (There's a timeout involved, which might presumably lead to timing issues if large texts are involved. Not sure if CTRL-A on large texts is such a big use case though?) Browsers behave in subtly different ways when it comes to the details. But overall, everything seems to work.

I tested with the standard test page from the `packages/text-annotator-react` (which includes the dummy popup with a text input field). On the following devices/browsers: 

- Mac (desktop, mouse/touchpad): Chrome, Firefox, Safari
- Mac (desktop, keyboard): Chrome, Firefox
- iPad: Chrome, Firefox, Safari
- Android: Chrome, Firefox, Samsung Internet (for whatever the hell that is :-)

packages/text-annotator/src/utils/device.ts
oleksandr-danylchenko, line 1: Maybe we can use a 3rd party library like https://www.npmjs.com/package/ua-parser-js? It's constantly maintained and will account for any future UA string shenanigans.

packages/text-annotator-react/src/TextAnnotatorPopup/isMobile.ts
oleksandr-danylchenko, line 1: https://github.com/recogito/text-annotator-js/pull/151/files#r1786028452

packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
```diff
@@ -69,38 +65,31 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
...
-      setOpen(false);
-    }
-  }, [isOpen, selectedKey]);
+    setOpen(selected.length > 0);
+  }, [selected.map(a => a.annotation.id).join('-')]);
 
   useEffect(() => {
     if (isOpen && annotation) {
+      // Extra precaution - shouldn't normally happen
+      if (!annotation.target.selector || annotation.target.selector.length < 1) return;
```
oleksandr-danylchenko, line 80: Nice addition ✔️ 
It's helpful when the consumer uses a custom plugin to filter out selectors. Which is exactly the case I had and not accounting for the missing `selector` shot me in a knee

packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
```diff
@@ -13,13 +16,12 @@ import {
...
 } from '@floating-ui/react';
 
-import { useAnnotator, useSelection } from '@annotorious/react';
-import type { TextAnnotation, TextAnnotator } from '@recogito/text-annotator';
-
 import './TextAnnotatorPopup.css';
 
 interface TextAnnotationPopupProps {
 
+  ariaCloseWarning?: string;
```
oleksandr-danylchenko, line 23: ✅

packages/text-annotator-react/test/App.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { FC, useCallback, useEffect } from 'react';
 import { AnnotationBody, Annotorious, useAnnotationStore, useAnnotator, useSelection } from '@annotorious/react';
-import { TextAnnotator, TextAnnotatorPopup, type TextAnnotationPopupContentProps } from '../src';
+import { TextAnnotator, TextAnnotatorPopup } from '../src';
 import { W3CTextFormat, type TextAnnotation, type TextAnnotator as RecogitoTextAnnotator } from '@recogito/text-annotator';
 
-const TestPopup: FC<TextAnnotationPopupContentProps> = (props) => {
+const TestPopup = (props) => {
```
oleksandr-danylchenko, line 6: Why did the type get removed? 🤔 
The `props.annotation` is still being accessed

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -212,23 +221,87 @@ export const SelectionHandler = (
...
+
+    const sel = document.getSelection();
+
+    if (sel?.isCollapsed) return;
+
+    // When selecting the initial word, Chrome Android fires `contextmenu` 
+    // before selectionChanged.
+    if (!currentTarget || currentTarget.selector.length === 0) {
+      onSelectionChange(evt);
+    }
```
oleksandr-danylchenko, line 251: Wow! Good to know!

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -212,23 +221,87 @@ export const SelectionHandler = (
...
+        selection.userSelect(currentTarget.annotation, cloneKeyboardEvent(evt));
+      }
+      
+      document.removeEventListener('selectionchange', onSelected);
+
+      // Sigh... this needs a delay to work. But doesn't seem reliable.
+    }, 100);
+
+    // Listen to the change event that follows
+    document.addEventListener('selectionchange', onSelected);
```
oleksandr-danylchenko, line 291: Do I understand correctly that with the delay we make the handler wait till the existing `selectionchange` listener processes the selection and create the target? So then we'll just persist it

packages/text-annotator/src/utils/device.ts
rsimon: Good point - I'll merge first, but will create an issue for this.

packages/text-annotator-react/test/App.tsx
rsimon: Hm, not sure. I re-added!

packages/text-annotator/src/SelectionHandler.ts
rsimon: So here's what I could figure out: if you hit `Cmd`+`A` it will take a while until the selection is actually done. Initially, I just waited for a while in the key listener (250ms) and then stored the current selection. But that seemed really brittle. (What if the selection hasn't happened yet, and `currentTarget` is still empty.)

Therefore, I implemented the above: 
- When `Cmd`+ `A` is pressed, I'm attaching a listener to listen for the `selectionchange` event.
- Assuming that the selectionchange event would fill the `currentTarget`, I would simply create the annotation, and remove the listener again.

It turned out that this wasn't sufficient. I still had to wait a bit more. I don't think it's event order (my listener getting triggered before the one that fills the `currentTarget`). Because if it were just about the order, a wait time of 1ms should be enough? But that didn't work.

packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
```diff
@@ -152,13 +139,12 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
             event
           })}
 
-          {/* It lets keyboard/sr users to know that the dialog closes when they focus out of it */}
-          <button className="popup-close-message" onClick={handleClose}>
-            This dialog closes when you leave it.
-          </button>
+          <span className="r6o-popup-sr-only" aria-live="assertive">
+            {props.ariaCloseWarning || 'This dialog will close when you leave it.'}
+          </span>
```
oleksandr-danylchenko, line 144: Unfortunately, that message won't be picked by the screenreader. Because `aria-live` announces only the changes in the content within the live area. Therefore, the message will simply be missed ;(

oleksandr-danylchenko: That's why I used the `@react-aria/live-announcer` package in https://github.com/recogito/text-annotator-js/pull/141. Furthermore, the live area won't even announce messages that are repeated! So I needed to trick it by inserting a random number of `&nbsp;` characters into it. 
[See the source](https://github.com/recogito/text-annotator-js/pull/141/files#diff-86cf4735dfec418174559a308bca544a80c6178258372614d602a699166b0602R27-R28)

rsimon: Ok, interesting. One thing I changed is that I made this a `<span>` rather than a button. Would a button make the message appear?

oleksandr-danylchenko: And because of those reasons I went with the button, recommended by a11y expect. Because it's an explicit "speedbump" that will require an action from a user

rsimon: Or does this mean we should skip this altogether? (In favour of #141)

oleksandr-danylchenko (commented):
The changes look great and clear, thank you 🔥 
The only feasible concern I have is losing the focus-out speedbump for the popup.

packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
oleksandr-danylchenko: Unfortunately, they are unrelated. In #141 I added a navigation message that will help users to know that they should press `Tab` to navigate to the popup. But it doesn't cover the focus-out speedbump. It was added in https://github.com/recogito/text-annotator-js/pull/118, [source](https://github.com/recogito/text-annotator-js/pull/118/files#diff-ad12777004d12b40f8adf25b1092fccc60c12ea2dd2d352ec0260079190780e8R155-R158)

oleksandr-danylchenko: > Would a button make the message appear?

Yeah, it would, because the button is natively focusable. That's why the `r6o-popup-sr-only` has those `&:focus` appearing styles: https://github.com/recogito/text-annotator-js/blob/8d113dc93167df8d48521f331649202bf6cf32a1/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.css#L20-L30

rsimon: Ok, sounds good. I'll revert to the button then. But I wonder if the default message should be modified accordingly. (It is still a button, so users with screen readers would also be able to 'click' it, right?) How about, simply 'Close this dialog'? Or 'Click to close this dialog. Leaving the dialog will close automatically.'?

rsimon: Or perhaps more compact: __Click or leave this dialog to close it.__

oleksandr-danylchenko: > Or 'Click to close this dialog. Leaving the dialog will close automatically.'?

I like this option ✅ 
It's pretty clear and fits all the purposes

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -142,21 +143,12 @@ export const SelectionHandler = (
...
 
+    /**
+     * During mouse selection on the desktop, annotation won't usually exist while the selection is being edited.
+     * But it will be typical during keyboard or mobile handlebars selection!
+     */
     if (store.getAnnotation(currentTarget.annotation)) {
       store.updateTarget(currentTarget, Origin.LOCAL);
-    } else {
-      // Proper lifecycle management: clear selection first...
-      selection.clear();
```
oleksandr-danylchenko, line 149: I think that we may need to restore the `selection.clear()` call if there's nothing to update. Because now the previous, discarded selection stays up until you create a brand-new one.

https://github.com/user-attachments/assets/cb0d6edc-5a48-400b-90f8-a61c41251200

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -212,23 +223,87 @@ export const SelectionHandler = (
...
+      const sel = document.getSelection();
 
       // Just a click, not a selection
       if (sel?.isCollapsed && timeDifference < CLICK_TIMEOUT) {
         currentTarget = undefined;
         clickSelect();
-      } else if (currentTarget && store.getAnnotation(currentTarget.annotation)) {
-        selection.userSelect(currentTarget.annotation, evt);
+      } else if (currentTarget && currentTarget.selector.length > 0) {
+        selection.clear();
```
oleksandr-danylchenko, line 233: Also... I'm not entirely sure why the _current selection_ gets cleared and then immediately reselected back 🤔

packages/text-annotator/src/SelectionHandler.ts
oleksandr-danylchenko: UPD: Submitted a PR - https://github.com/recogito/text-annotator-js/pull/155

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -177,7 +171,23 @@ export const SelectionHandler = (
...
+      store.addAnnotation({
+        id: currentTarget.annotation,
+        bodies: [],
+        target: currentTarget
+      });
+    }
+  }
+
   const onPointerUp = (evt: PointerEvent) => {
+    if (isContextMenuOpen) return;
```
oleksandr-danylchenko, line 189: Could you please clarify why the context menu guard was added for the `pointer` event handlers? With it in place, it takes 2 taps to dismiss the annotation selection. The first tap is needed to dismiss the context menu and the second one is to dismiss the annotation itself.

However, the annotation popup gets dismissed on the first tap too due to the `outside-press` reason! 
image

That makes the annotation selection look like it's lagging behind 🤔 

https://github.com/user-attachments/assets/d9e3526c-dedb-4582-9a3c-a055968e79a3

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -166,6 +158,8 @@ export const SelectionHandler = (
    * to the initial pointerdown event and remember the button
    */
   const onPointerDown = (evt: PointerEvent) => {
+    if (isContextMenuOpen) return;
```
oleksandr-danylchenko, line 161: Could you please clarify why the context menu guard was added for the `pointer` event handlers? With it in place, it takes 2 taps to dismiss the annotation selection. The first tap is needed to dismiss the context menu and the second one is to dismiss the annotation itself.

However, the annotation popup gets dismissed on the first tap too due to the `outside-press` reason! 
image

That makes the annotation selection look like it's lagging behind 🤔 

https://github.com/user-attachments/assets/d9e3526c-dedb-4582-9a3c-a055968e79a3

oleksandr-danylchenko, line 161: Hello, @rsimon 👋🏻 

Could you please clarify why the context menu guard was added for the `pointer` event handlers? With it in place, it takes 2 taps to dismiss the annotation selection. The first tap is needed to dismiss the context menu and the second one is to dismiss the annotation itself.

However, the annotation popup gets dismissed on the first tap too due to the `outside-press` reason! 
image

That makes the annotation selection look like it's lagging behind 🤔 

https://github.com/user-attachments/assets/d9e3526c-dedb-4582-9a3c-a055968e79a3

oleksandr-danylchenko: I see that it was added under the `Minor fix` commit https://github.com/recogito/text-annotator-js/commit/fb3d70bedba11660d64b479460094e951e530d38. However, it doesn't carry enough context 😅

oleksandr-danylchenko: Aha... The "bug" manifested itself only when I ran the React example w/o the `isContextMenu` check: 
image

oleksandr-danylchenko: Oh wait... The same exception exists even on the `main` branch when I run the React example on Android!

Issue #153, opened 2024-10-03
Controlling selection behaviour


Aierie:
Hello! Wondering about controlling selection behaviour from userland. 

I am making an app that has different behaviours on selection when different keys are pressed, and needs to be able to conditionally select a group of related annotations when one of them are clicked, and toggle selection state of a single annotation without interfering with other annotations. Initially, looking at the interface of `clickAnnotation` event [here](https://github.com/annotorious/annotorious/blob/7f712d213412517a1d3a9ef40c4787abfa7927b3/packages/annotorious-core/src/lifecycle/LifecycleEvents.ts#L5), I thought I would be able to retrieve whether modifier keys are pressed from the original event. However, it seems like `clickAnnotation` is not implemented for `@recogito/text-annotator`.

I then tried listening for DOM click events and checking whether they were on a range of an annotation:

```js
document.querySelector('#annotator-target-example-6').addEventListener('click',(e) => {
	let range,textNode,offset;
	if (document.caretPositionFromPoint) {
		range = document.caretPositionFromPoint(
			e.clientX,
			e.clientY,
		);
		textNode = range.offsetNode;
		offset = range.offset;
	} else if (document.caretRangeFromPoint) {
		// Use WebKit-proprietary fallback method
		range = document.caretRangeFromPoint(e.clientX, e.clientY);
		textNode = range.startContainer;
		offset = range.startOffset;
	} else {
		throw new Error('document.caretPositionFromPoint and document.caretRangeFromPoint are not supported');
	}
	let annotations = annotator.getAnnotations();
	let annotationId = annotations.map((v) => [v.id, v.target.selector[0].range]).find(([k, r]) => { return r.isPointInRange(textNode, offset); })[0];
	if (!annotationId) return;
	if (e.altKey) {
		let selectedIds = new Set(annotator.getSelected().map((a) => a.id));
		if (selectedIds.has(annotationId)) {
			selectedIds.delete(annotationId);
		} else {
			selectedIds.add(annotationId);
		}
		annotator.setSelected(
			[...selectedIds]
		)
	} else {
		annotator.setSelected(
			[
				annotationId
			]
		)
	}
});
```

But it seems like internally, even with `userSelectAction` set to `UserSelectAction.NONE`,  `TextAnnotatorState` is calling `userSelect` here:

https://github.com/recogito/text-annotator-js/blob/ead48404841271de4f3df97da8f2d368736956fb/packages/text-annotator/src/SelectionHandler.ts#L197

[...]

PR #154, opened 2024-10-04
prevents select on click when userSelectAction is none

Aierie:


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -193,6 +193,10 @@ export const SelectionHandler = (
 
     // Logic for selecting an existing annotation
     const clickSelect = () => {
+      if (selection.userSelectAction === UserSelectAction.NONE) {
+        return;
+      }
+
```
oleksandr-danylchenko, line 199: That logic is already covered in the parent `@annotorious/annotorious` package. Even though the `selection.userSelection` will be invoked, the actual selection state will always remain empty. [See this source](https://github.com/annotorious/annotorious/blob/03ee23fa905a3fc448b4aa51d769c15ccb31c93a/packages/annotorious-core/src/state/Selection.ts#L69-L79)

oleksandr-danylchenko: Also, we cannot compare the `userSelectAction` with the enum value directly, because the `userSelectAction` can also be an expression: `((a: T) => UserSelectAction)`

Aierie: Yes - what I'm trying to achieve is to control the selection myself, described in #153

Aierie: I'll work on the `userSelectAction` expression, thanks! I just realised this - sorry, am pretty new here.

rsimon: Good point. Yes, indeed the core package already checks this against the current `userSelectionAction` (value or function).

Aierie: > That logic is already covered in the parent @annotorious/annotorious package.

The logic in the parent package seems to set the selection to an empty array if `userSelectAction` is `UserSelectAction.NONE`. Is this intended behaviour?

rsimon: Yes that's as intended. The three settings are:

- `EDIT` - a hint to the UI that selecting should make the annotation shape editable. (Only supported for image annotations so far.)
- `SELECT` - should make selected annotations surface in the selection event, but not make it editable.
- `NONE` - should make annotations inert. Clicking does nothing, not even fire selection events.

Aierie: `NONE` right now doesn't make annotations inert though. It removes programmatic selections

Aierie: Yes, I'm trying to do multi-selection behaviour. The selection API does support a list of annotations, but I have to set `userSelectAction` to `NONE`. And when I do that, clicking on any annotation will clear the list of annotations that my program has selected

rsimon: > I guess - my expectation of inertness is that this default clause does nothing.

I guess that's something that might be worth discussing. In principle, my view is: clicking an inert annotation would be considered the same as clicking on empty space, which is supposed to clear the selection. What's __not__ covered in the current code, of course, is that you might click an inert annotation that overlaps a non-inert one. In this case, I'd expect the non-inert one to get selected (and the inert one to get ignored). The current code, however, would simply clear the selection as you say.

Aierie:
Thanks, will share once I've cleaned it up enough to use! It is a tool for doing bottom-up qualitative data synthesis that puts a lot of importance on the raw quotes, following [Indi Young](https://indiyoung.com/)'s method.

I was originally hacking around with RecogitoJS, but realised the successor checked these boxes:

- [x] Better rendering to avoid awkward UI when applying certain custom styles to annotation spans
- [x] Undo/redo
- [x] More control over selection

So thanks! This saves me a bunch of work while making the code cleaner. I'll continue using `@recogito/text-annotator-js`, but will probably store selection state as a property in the annotation for now.

rsimon:
Sounds great, thanks!

PR #155, opened 2024-10-10
Added missing previous selection discard upon the new selection change

oleksandr-danylchenko:
## Issue
Initially reported - https://github.com/recogito/text-annotator-js/pull/151#discussion_r1795661665

> We need to restore the selection.clear() call if there's nothing to update. Because now the previous, "discarded selection" unexpectedly stays up until you create a brand-new one.
>
> https://github.com/user-attachments/assets/cb0d6edc-5a48-400b-90f8-a61c41251200
>
>*in my testing env, the annotations that don't have either a highlight or a note assigned get automatically garbage collected when they get unselected

Can be reproduced in the playground: 

https://github.com/user-attachments/assets/81bf9fc7-380d-4438-a81e-d0d6365b251d

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -209,7 +209,7 @@ export const SelectionHandler = (
         if (selected.length !== 1 || selected[0].id !== hovered.id) {
           selection.userSelect(hovered.id, evt);
         }
-      } else if (!selection.isEmpty()) {
+      } else {
         selection.clear();
       }
     };
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -149,6 +149,9 @@ export const SelectionHandler = (
      */
     if (store.getAnnotation(currentTarget.annotation)) {
       store.updateTarget(currentTarget, Origin.LOCAL);
+    } else {
+      // Proper lifecycle management: clear the previous selection first...
+      selection.clear();
     }
   });
 


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -206,7 +209,7 @@ export const SelectionHandler = (
         if (selected.length !== 1 || selected[0].id !== hovered.id) {
           selection.userSelect(hovered.id, evt);
         }
-      } else if (!selection.isEmpty()) {
+      } else {
         selection.clear();
       }
```
oleksandr-danylchenko, line 214: The `isEmpty` check is redundant because such a case is already handled on the a9s side. The operation will be no-op if the selection is already clear. [Source](https://github.com/annotorious/annotorious/blob/03ee23fa905a3fc448b4aa51d769c15ccb31c93a/packages/annotorious-core/src/state/Selection.ts#L45-L49)

rsimon:
Thanks - yes, that makes sense. The previous selection should clear as soon as you start the next, not after you complete it.

oleksandr-danylchenko:
### I FOUND A REGRESSION IN THIS CHANGE
It breaks the existing annotations selection on Android! For an unknown reason, the order of the events is different on Android compared to the Mac or iOS when a user clicks over the text!
| Mac | Android |
|--------|--------|
| image | image  | 

So on Android when the annotation gets "clicked" in the `pointerup` handler - the selection gets immediately cleared in the `selectionchange` handler 😭. Which makes the selection of the existing annotations impossible.

oleksandr-danylchenko:
Another mobile quirk!
On `iOS` the ordering is the same as on Android - the `selection` comes after the `pointer` events. BUTT! It fluctuates depending on where and how fast you click!
When you click over an annotation quickly - only `pointerdown`/`pointerup` are dispatched. But when slowly - all 4. When I tap over the textual content - all 4.

https://github.com/user-attachments/assets/cac92b06-5989-46d4-a7c8-9798723ea18f

PR #157, opened 2024-10-14
⚠️ Fixed false-negative `.not-annotatable` presence check ⚠️

oleksandr-danylchenko:
## Issue
Previously, only the `parentElement` of the pointer event target was checked for the `not-annotatable` class presence. Unfortunately, such a check misses a case when the element itself can be a target and have the `not-annotatable` class assigned!
image
## Changes Made
Added a shared `isNotAnnotatable` utility that starts the check directly from the element when possible.


packages/text-annotator/src/utils/getAnnotatableFragment.ts
oleksandr-danylchenko, line 1: This utility wasn't used anywhere and its functionality is replicated in the: https://github.com/recogito/text-annotator-js/blob/775f6125f363f76b2aeb27d2bcb289d8f74a1c4b/packages/text-annotator/src/utils/splitAnnotatableRanges.ts#L77-L81

oleksandr-danylchenko:
!! Our QC team discovered the following exception can occasionally happen when testing the app on iOS devices:
image

oleksandr-danylchenko:
❗❗  Our QC team discovered the following exception can occasionally happen when testing the app on iOS devices:
image
It happens when buttons are clicking within the modal that is marked as `not-annotatable`. Investigating it now 🔎

oleksandr-danylchenko:
UPD: Addressed the issue in https://github.com/recogito/text-annotator-js/pull/164

PR #158, opened 2024-10-14
Fixed overeager popup closure on touch devices

oleksandr-danylchenko:
## Issue
When a user scrolls upon the popup opening - the latter gets closed instead of being scrolled along! The issue is the [`outsidePressEvent`](https://floating-ui.com/docs/useDismiss#outsidepressevent) that's `pointerdown` by default. Therefore, it would immediately dismiss the popup when a user touches their screen. Instead, we can use a bit more touch-friendly `mousedown`: 
```ts
useDismiss(context, {
  // Eager on both mouse + touch input.
  outsidePressEvent: 'pointerdown',
  // Eager on mouse input; lazy on touch input.
  outsidePressEvent: 'mousedown',
  // Lazy on both mouse + touch input.
  outsidePressEvent: 'click',
});
```
So when a user touches a screen and drags the popup won't be dismissed, only when the user touches and releases almost immediately.

### DIFF
--- a/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
+++ b/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
@@ -111,12 +111,6 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
     };
   }, [update]);
 
-  // Prevent text-annotator from handling the irrelevant events triggered from the popup
-  const getStopEventsPropagationProps = useCallback(
-    () => ({ onPointerUp: (event: PointerEvent<HTMLDivElement>) => event.stopPropagation() }),
-    []
-  );
-
   // Don't shift focus to the floating element if selected via keyboard or on mobile.
   const initialFocus = useMemo(() => {
     return (event?.type === 'keyup' || event?.type === 'contextmenu' || isMobile()) ? -1 : 0;
@@ -136,8 +130,7 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
           className="a9s-popup r6o-popup annotation-popup r6o-text-popup not-annotatable"
           ref={refs.setFloating}
           style={floatingStyles}
-          {...getFloatingProps()}
-          {...getStopEventsPropagationProps()}>
+          {...getFloatingProps(getStopEventsPropagationProps())}>
           {props.popup({
             annotation: selected[0].annotation,
             editable: selected[0].editable,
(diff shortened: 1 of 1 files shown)


packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx
```diff
@@ -138,8 +130,7 @@ export const TextAnnotatorPopup = (props: TextAnnotationPopupProps) => {
           className="a9s-popup r6o-popup annotation-popup r6o-text-popup not-annotatable"
           ref={refs.setFloating}
           style={floatingStyles}
-          {...getFloatingProps()}
-          {...getStopEventsPropagationProps()}>
+          {...getFloatingProps(getStopEventsPropagationProps())}>
```
oleksandr-danylchenko, line 133: FloatingUI mandates that any custom event listeners should be propagated through the `getReferenceProps`/`getFloatingProps`, [source](https://arc.net/l/quote/cqodxysn): 
> This is because your handler may be either overwritten or overwrite one of the Hooks’ handlers. More event handlers may also be added in future versions.

oleksandr-danylchenko:
We shouldn't close the `popup` on the pointer interactions at all! That should be controlled by the `SelectionHandler`! Otherwise, we risk introducing a conflicting race condition between the handler's "annotation click" and the previous annotation popup closure. So clicking from annotation to annotation would have a jittering effect

rsimon:
I was thinking about this, and had my doubts as well. I think the `ESC` key behavior was useful. But agree it would be cleaner to handle this on the `SelectionHandler` level.

oleksandr-danylchenko:
The only thing that can be improved a bit in the `SelectionHandler` is the support for scrolling on the touch devices. Currently, when the popup is opened and you scroll a bit, that gets considered as a click, and the popup is dismissed. However, the native context menu is still present. So we may wanna narrow the "click" check down a bit to compare the `down`/`up` coordinates.

oleksandr-danylchenko:
> when the popup is opened and you scroll a bit, that gets considered as a click, and the popup is dismissed. However, the native context menu is still present

The issue is resolved by moving the `setOpen` under the `escape-key` & `focus-out` reasons. Then, the `outside-click` won't accidentally dismiss the popup, but put the responsibility onto the `SelectionHandler`.
https://github.com/recogito/text-annotator-js/blob/69be7ea1b6a8be3dcd4e02e7803179c6a37a4f6d/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx#L52-L57

> So we may wanna narrow the "click" check down a bit to compare the down/up coordinates.

We don't need to do that, because the `pointerup` is not even dispatched when a user scrolls on the touch device even slightly! Only the `pointerdown`. Therefore, the "click" won't even be processed!
image

Issue #159, opened 2024-10-15
Dismissing the selection on Android takes 2 taps


oleksandr-danylchenko:
## Issue
Certain reasons filter the annotation selection dismissal: https://github.com/recogito/text-annotator-js/blob/775f6125f363f76b2aeb27d2bcb289d8f74a1c4b/packages/text-annotator-react/src/TextAnnotatorPopup/TextAnnotatorPopup.tsx#L52-L57
So when a user on Android presses outside of the popup area - the selection doesn't get dismissed. Instead, they need to press twice. The first time the focus is placed outside of the popup, and the second time the tap is recognized as a "click" over a not-selectable area by the `SelectionHandler`.

https://github.com/user-attachments/assets/ad7e5d97-ca3d-437a-a641-91ee685a458d

## Suggested Changes
~Remove the `reason` filtering and always dismiss the annotation selection when the popup is about to be closed. As the closure reason usually will be the `outside-press`, `focus-out`, and `escape-key`.~

---

The issue turned out to be related to the `isCollapse` still being `false` even though a user tapped outside of the active selection...
See - https://github.com/recogito/text-annotator-js/issues/136#issuecomment-2413773804 & https://github.com/recogito/text-annotator-js/pull/163

---

Another issue is that the `pointerup`/`pointerdown` events handling was gated by the `isContextMenu` flag. See - https://github.com/recogito/text-annotator-js/pull/167
UPD: ✅

oleksandr-danylchenko:
UPD: The issue turned out to be related to the `isCollapse` still being `true` even though a user tapped outside of the active selection...
See - https://github.com/recogito/text-annotator-js/issues/136#issuecomment-2413773804

PR #160, opened 2024-10-15
#159 Removed annotation selection dismissal filtering by a `reason`

oleksandr-danylchenko:
## Issue - https://github.com/recogito/text-annotator-js/issues/159


oleksandr-danylchenko:
We still need the `reason` filtering, because the popup shouldn't be dismissed, or, even worse, annotation selection shouldn't get cleared on `outside-click`. That will lead to the race condition between the `SelectionHander` and `FloatingUI` who will cancel the selection first.

PR #161, opened 2024-10-15
Simplified popup opening check with `isRevived`

oleksandr-danylchenko:
## Issue
The popup opened state check is almost a duplicate of the existing `isRevived` utility. It already does the check for the ranges presence and whether they are not collapsed.


rsimon:
Sounds good!

Issue #136, opened 2024-08-14
Annotation isn't dismissed clicking an empty space, covered by the native selection range ⚠️


oleksandr-danylchenko:
## Background
While handling the `pointerup` event, checking the `document.getSelection().isCollapsed` property can behave in 2 ways:
1. You made a selection and clicked on it. Then, the `isCollapsed` is `false` even though the native highlight disappears! ⚠️
2. You made a selection and clicked somewhere else. Then, the `isCollapsed` is expectedly `true`, matching the native highlight disappearance.

https://github.com/user-attachments/assets/919f9e25-0852-4aec-85fa-983532243060

## Issue

In the `SelectionHander` the annotation selection gets dismissed only when the `pointerdown` + `pointerup` combination is recognized as the "click" and there's nothing "hovered" (lines 152-155 & 144-146):
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/SelectionHandler.ts#L149-L157
https://github.com/recogito/text-annotator-js/blob/e6d3f33339ab1f4cd2ccb6b77624c20f12841ee4/packages/text-annotator/src/SelectionHandler.ts#L134-L147

However, as the `document.getSelection().isCollapsed` is `false` when you click over the native selection range, _the annotation selection sometimes doesn't get dismissed_!
Instead, _the annotation gets selected another time_, because the `currentTarget` is still present!

Alternatively, we expect that the annotation selection will behave like the native selection highlight in all cases. When you click on an empty space - it should be dismissed.

## Issue Examples

https://github.com/user-attachments/assets/aee59f76-890b-4571-9614-03ca3bb051e2

https://github.com/user-attachments/assets/2997c3f1-5da3-4b15-935c-b9dad0cfd587

## Suggested Changes
The root of the issue is the `document.getSelection().isCollapsed` that is "lagging behind" on the `pointerup` event handling.

After I added a 0ms. timeout for the reading, I started getting more adequate value consistently. It's always `true` either when you click on the selected range or some other content.

https://github.com/user-attachments/assets/35ef211b-06cb-4669-a50f-d6e0764aec68

In that way, the "click" will always get processed consistently and the annotation selection will get dismissed when you didn't click on the highlight explicitly.

rsimon:
Just adding that this should now also work in the [revised-selection-behavior](https://github.com/recogito/text-annotator-js/tree/revised-selection-behavior) branch.

[...]

PR #162, opened 2024-10-15
Increased the `collapsed` check timeout on `pointerup`

oleksandr-danylchenko:
## Issue
Source - https://github.com/recogito/text-annotator-js/issues/136#issuecomment-2413773804
> Unfortunately, I noticed that the "lagging behind `isCollapsed` property" is still a thing for touch devices! When I tap somewhere _around_ the selection - the `isCollapsed` is still `true` when the `onPointerUp` gets handled, even within the `setTimeout`. 
image
> That creates a weird effect that makes a user _tap twice_ to dismiss the selection. Similar to the one we had in the demos above ;(

## Changes Made
Source - https://github.com/recogito/text-annotator-js/issues/136#issuecomment-2413810386
> Incredibly... increasing the `setTimeout`'s timeout from the default 0 to 1 helps! It starts reporting the `isCollapsed` properly again!
> image

UPD: Turned out that 1ms. isn't always enough, so I switched to the 5ms. delay instead


oleksandr-danylchenko:
I think that approach should be more flexible here...
I discovered that on `iOS` it takes up to **_`40ms`_** to properly report that the selection range is collapsed even when a user clicks beyond the existing selection! But if we decide to apply such a timeout universally - it'll make user experience on desktops and Androids slower in general.
Instead, we may wanna have a "polling" for the range collapse state 🤔. If the `timeDifference` is smaller than the `CLICK_TIMEOUT` and the range isn't collapsed - we start polling it for up to 50ms. (to be extra cautious). So if the browser changes its state quicker - it won't be held off by a "blind" timeout.

PR #163, opened 2024-10-15
⚠️ Fixed accessing an obsolete `isCollapsed` value on `pointerup` handling ⚠️

oleksandr-danylchenko:
## Issue
Source - https://github.com/recogito/text-annotator-js/issues/136#issuecomment-2413773804
> Unfortunately, I noticed that the "lagging behind `isCollapsed` property" is still a thing for touch devices! When I tap somewhere _around_ the selection - the `isCollapsed` is still `false` when the `onPointerUp` gets handled, even within the `setTimeout`. 
image
> That creates a weird effect that makes a user _tap twice or even more on iOS_ to dismiss the selection.

Instead, I introduced "polling" for the range's collapsed state! If the `timeDifference` is smaller than the `CLICK_TIMEOUT` and the `range` isn't collapsed - we start polling it for up to `50ms.` (to be extra cautious). So if the browser changes its state quicker - it won't be held off by a "blind" timeout.


oleksandr-danylchenko:
Unfortunately, it still takes 2 taps on Android to dismiss the selection 🤔

rsimon:
Done! I'm getting the following during build though:

```
No name was provided for external module "poll" in "output.globals" – guessing "poll".
```

Not sure if the new `poll` dependency is UMD compatible?

oleksandr-danylchenko:
Although, _I don't think it'll work anyway_ in the browser if only the R6O dependency is imported ⚠️
That's because the `global` simply won't have the `.UUID`, `. AnnotoriousCore`, etc. populated. As those packages don't inject themselves into a global context.

PR #166, opened 2024-10-29
Fixed false-positive `open` setting

oleksandr-danylchenko:
## Issue
This PR is a follow-up on https://github.com/recogito/text-annotator-js/pull/161 where I accidentally introduced a false-positive setting for the `open` state of the popup. In the current logic, despite the `annotation?.target.selector` being an empty array - it'll still pass the `isRevived` check due to the `.every` nature!
https://github.com/recogito/text-annotator-js/blob/69be7ea1b6a8be3dcd4e02e7803179c6a37a4f6d/packages/text-annotator/src/utils/isRevived.ts#L3-L4
image

Therefore:

image


oleksandr-danylchenko:
The issue above is nonexistent in https://github.com/recogito/text-annotator-js/pull/123 though. There I added the `r.state.store.getAnnotationBounds` usage to obtain the selection range dimensions. Therefore, if the `selector` is empty - it'll just return `undefined`. And on the 2nd+ hook re-run, the dimensions will be properly obtained ✅

PR #170, opened 2024-11-01
Added W3C `scope` property support parsing/serialization

oleksandr-danylchenko:
## Issue
We have a need to associate a certain W3C `target` with a certain covered "resource"/"element" on a page. To not create custom properties for the W3C model, like `element_family_id`, we decided to go with the existing [`scope` property](https://www.w3.org/TR/annotation-model/#scope-of-a-resource).
> It is sometimes important for an Annotation to capture the context in which it was made in terms of the resources that the annotator was viewing or using at the time

However, as for the `styleClass` (https://github.com/recogito/text-annotator-js/pull/72), it's not enough to declare the type for W3C/internal model locally. It needs to be explicitly processed by the `W3CTextFormatAdapter`.

### DIFF
--- a/packages/text-annotator/src/model/core/TextAnnotation.ts
+++ b/packages/text-annotator/src/model/core/TextAnnotation.ts
@@ -24,6 +24,8 @@ export interface TextSelector {
 
   range: Range;
 
-  offsetReference?: HTMLElement
+  offsetReference?: HTMLElement;
+
+  scope?: string;
 
 }
--- a/packages/text-annotator/src/model/w3c/W3CTextAnnotation.ts
+++ b/packages/text-annotator/src/model/w3c/W3CTextAnnotation.ts
@@ -14,6 +14,8 @@ export interface W3CTextAnnotationTarget extends W3CAnnotationTarget {
 
   styleClass?: string;
 
+  scope?: string;
+
 }
 
 /**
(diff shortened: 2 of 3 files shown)


oleksandr-danylchenko:
> They'll have to be handled by the crosswalk adapter only

But how would the crosswalk adapter know about the existence of such properties for serialization if they are not present in the Core type?

rsimon:
You can always have custom properties if you want them. It's just JS after all. But I think it defeats the point of having an external format if every extension is mirrored 1:1 in the internal model (which doesn't use that extension).

oleksandr-danylchenko:
So would you recommend setting up a local W3C adapter that respects those properties? It seems like quite a bit of a hassle that will divert the consumer's code from the generic adapter present in the annotator already. 🤔

rsimon:
I think it's totally fine for the W3CAdapter to have and handle those props. But the core interface should not declare them.

For the crosswalks, the adapter would just have to use statements like `if 'styleClass' in...`, or use spread operators to copy all props wherever it makes sense, I guess?

rsimon:
I added a change, removing `scope` from the core interface. At the moment, there's one remaining `@ts-ignore`. I assume this could be avoided by re-typing the the internal "side" of the adapter from `TextAnnotation` to `ScopedTextAnnotation` or something. I didn't dive into this now, but given that the internal side is also generic now, this should be possible? (And the `ScopedTextAnnotation` would be in the `w3c` folder, keeping it separate from the core.)

Issue #172, opened 2024-11-04
Selecting on touch devices causes highlight flickering when move down rapidly 


oleksandr-danylchenko:
## Issue
I noticed that when the `TextAnnotatorPopup` is rendered and I move select text down rapidly - the finger can go over the floating element and cause the highlight flickering:

https://github.com/user-attachments/assets/ebdb8383-c4f1-4f49-859e-1ffd7d1818b7

However, it doesn't happen when I remove the popup rendering: 

https://github.com/user-attachments/assets/9b208e65-39a2-49e6-a544-23018b4c66ed

---

In the real app on iOS it looks even more flickery when the selection goes over the headings: 

https://github.com/user-attachments/assets/3f60494a-1b35-4467-9339-5a73dd2757f3

rsimon:
Yes, I saw this, too. But haven't investigated further yet. 

The selection clearly breaks as it goes from the selection start (on the text) to somewhere on the portal popup. I believe the popup is already marked as `not-annotatable`? Originally, I had the popup oriented into the upwards direction. In practice, that helps a bit since people will _usually_ want to select downward. (Although there's no guarantee, of course.) But because of the native context menu popup, downwards orientation really is the only option. Not sure what a good way around this would be...

oleksandr-danylchenko:
>  But haven't investigated further yet

No problem, I'm on it now.

Here's the selection change event `range` breakdown:
![image](https://github.com/user-attachments/assets/c0123b73-451c-4f92-acfd-f032d4487b69)

When the jump occurs, the `SelectionHandler` thinks that you selected down to the end of the annotatable container. However, the range updates back to the "normal" value almost instantly.

rsimon:
Thanks!! Hm, seems like the selection goes to the popup, which itself is not annotatable. Therefore, the clipping will select __up to__ the popup which, because the popup is portalled, is the end of the page?

oleksandr-danylchenko:
> up to the popup which, because the popup is portalled, is the end of the page?

Right! It's exactly at the end of the page: 
![image](https://github.com/user-attachments/assets/6767dc04-f6ef-40d6-ad69-bc1b701627c7)

Therefore, the `SelectionHandler` just clamps the selection to the end of the annotatable container

oleksandr-danylchenko:
It seems to be a common issue in the annotating solutions though 😅 
[...]

PR #173, opened 2024-11-04
#172 Added `isIdling` state to the `SelectionHandler`

oleksandr-danylchenko:
## [Issue](https://github.com/recogito/text-annotator-js/issues/172)


oleksandr-danylchenko:
@rsimon, here's an example of how the `isIdling` state can look like for the `SelectionHandler`. I intentionally chose to store _just a single flag_ instead of some kind of `enum`, because that's the only state we're interested in for now. So it's more direct and takes less maintenance effort.

PR #174, opened 2024-11-06
#172 Added popup hiding when the annotation's target is actively updated

oleksandr-danylchenko:
## Issue - https://github.com/recogito/text-annotator-js/issues/172

> I noticed that when the `TextAnnotatorPopup` is rendered and I move to select text down rapidly - the finger can go over the floating element and cause the highlight flickering:
>
> https://github.com/user-attachments/assets/ebdb8383-c4f1-4f49-859e-1ffd7d1818b7

That's because the selection can accidentally go over the popup element that's portalled out of the annotatable content container at the bottom. So the selection just glues to its edge for a moment.

## Changes Made
I added an idling listener that renders the popup only when there are no updates to the annotation's `target` for some time. The "update" is recognized when the `quote` updates. _Otherwise, the popup shouldn't be concerned about the updates for the rest of the properties._
That prevents an accidental overlap with the user's touch and repositioning on the selection correction [over the whitespace](https://github.com/annotorious/annotorious/pull/467#issuecomment-2459832369).

---

I also tried another alternative - making the popup act like the native context menu. However, its tracking turned out to be [super unreliable and complex](https://github.com/annotorious/annotorious/pull/467#issuecomment-2459832369). So I retreated to a more pragmatic option.

###### (Soomo Staged)


oleksandr-danylchenko:
*The changes in this PR passed the QC on our side and user can no longer react to the popup during the selection ✔️

packages/text-annotator-react/src/hooks/useAnnotationQuoteIdling.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const hasTargetQuoteChanged = (oldValue: TextAnnotationTarget, newValue: TextAnnotationTarget) => {
+  const { selector: oldSelector } = oldValue;
+  const oldQuotes = oldSelector.map(({ quote }) => quote);
+
+  const { selector: newSelector } = newValue;
+  const newQuotes = newSelector.map(({ quote }) => quote);
+
+  return dequal(oldQuotes, newQuotes);
+};
```
oleksandr-danylchenko, line 66: Inspired by the A9S's `hasTargetChange` method, [source](https://github.com/annotorious/annotorious/blob/8eaf2bb53534ff6336aad16ff18d3e3dac260894/packages/annotorious-core/src/utils/diffAnnotations.ts#L24-L40)

packages/text-annotator-react/src/hooks/useAnnotationQuoteIdling.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const hasTargetQuoteChanged = (oldValue: TextAnnotationTarget, newValue: TextAnnotationTarget) => {
+  const { selector: oldSelector } = oldValue;
+  const oldQuotes = oldSelector.map(({ quote }) => quote);
+
+  const { selector: newSelector } = newValue;
+  const newQuotes = newSelector.map(({ quote }) => quote);
+
+  return !dequal(oldQuotes, newQuotes);
+};
```
oleksandr-danylchenko, line 66: Inspired by the A9S's `hasTargetChange` method, [source](https://github.com/annotorious/annotorious/blob/8eaf2bb53534ff6336aad16ff18d3e3dac260894/packages/annotorious-core/src/utils/diffAnnotations.ts#L24-L40)

Issue #169, opened 2024-11-01
Releasing `Shift` or "handlebar" over the same selection leads to redundant `updateTarget` calls


oleksandr-danylchenko:
## Issue
In the new `SelectionHandler` such events as "releasing the Shift" or "releasing the handlebar" are considered as "selection finishing events". In both cases, the `upsertCurrentTarget` gets called almost immediately. However, a user can simply press Shift or tap over the handlebar, which will also lead to redundant `updateTarget` calls upon the initial creation.

### Desktop

https://github.com/user-attachments/assets/d8362214-20c1-4e79-b4a1-6994bfc7aa63

### Mobile

https://github.com/user-attachments/assets/cc2614c4-c239-4d27-8f90-0682b503733f

![image](https://github.com/user-attachments/assets/ea72e1ae-82e4-4340-a5a5-dabab63fb1f8)

PR #175, opened 2024-11-06
#169 Fixed repeated updates for the unchanged target

oleksandr-danylchenko:
The reason for such behavior was a missing check of whether the target that's about to be updated is different from the existing one.

## Changes Made
I added the `updated` property comparison between the existing annotation target and the `currentTarget`. That effectively filters out repeated `Shift` key releases or taps over the handlebar, because there's no `selectionchange` event emitter that solely leads the `updated` prop change.

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -169,20 +169,6 @@ export const SelectionHandler = (
     isLeftClick = lastDownEvent.button === 0;
   };
 
-  // Helper
-  const upsertCurrentTarget = () => {
-    const exists = store.getAnnotation(currentTarget.annotation);
-    if (exists) {
-      store.updateTarget(currentTarget);
-    } else {
-      store.addAnnotation({
-        id: currentTarget.annotation,
-        bodies: [],
-        target: currentTarget
-      });
-    }
-  }
-
   const onPointerUp = (evt: PointerEvent) => {
     if (isNotAnnotatable(evt.target as Node) || !isLeftClick) return;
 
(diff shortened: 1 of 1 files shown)


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -317,6 +305,29 @@ export const SelectionHandler = (
...
+        bodies: [],
+        target: currentTarget
+      });
+      return;
+    }
+
+    const { target: { updated: existingTargetUpdated } } = existingAnnotation;
+    const { updated: currentTargetUpdated } = currentTarget;
+    if (
+      !existingAnnotation ||
```
rsimon, line 323: I think this one is redundant?

oleksandr-danylchenko: Hmm, I'm not sure that it's "redundant". As the `update` call was extracted into the `upsert` in the selection revision PR, https://github.com/recogito/text-annotator-js/pull/151, [source](https://github.com/recogito/text-annotator-js/pull/151/files#diff-1e072053802159d91da89e331e1fe166da3cb395a3fc6ab3d6317b0510ea17e0R174-R186). So there might was a reason for its existence 🤔

oleksandr-danylchenko: And if the `update` is needed in the `upsert` - comparing the `updated` timestamp is the key to preventing redundant calls to the `store.observe`

rsimon: Sorry, that wasn't clear from my side. I meant just the '!existingAnnotation' specifically. existingAnnotation is always true at this point unless I'm missing something.

oleksandr-danylchenko: Right... 
From the runtime perspective the value should exist at the point. By TS doesn't know about it due to the `updated?: Date` definition. So if we would have the strict null check enabled - it would complain that `existingTargetUpdated < currentTargetUpdated` would run on possibly `undefined` value. So I added the `!existingAnnotation` out of habbit

oleksandr-danylchenko: OHHHHH! I was meant to be the: 
```suggestion
      !existingTargetUpdated ||
```

rsimon: Glad I asked :-)

rsimon:
Thanks for this!!

PR #177, opened 2024-11-08
Simplified `arrow` middleware definition

oleksandr-danylchenko:
## Issue
In https://github.com/oleksandr-danylchenko/text-annotator-js/commit/ce7014d6f74d929488cb0c277f7015a5abc90b0e#r148871559 I spotted that the `arrow` middleware definition can be simplified to its default form. That should be safe for the floating itself because the `arrow` is a ["Data Middleware"](https://arc.net/l/quote/likqoxzb) that only populates the `context` with the positioning props, but doesn't change the behavior.

## Demo

https://github.com/user-attachments/assets/ab1dab97-7e4e-41e5-8737-66a55e218372


rsimon:
Tested the same thing in Annotorious - yep, works :-) Thanks!

PR #178, opened 2024-11-11
New `selectionMode` init prop

rsimon:
## In this PR

This PR introduces a new init property for the Text Annotator, `selectionMode`, which defines the behavior when the user selects an existing annotation that overlaps other annotations (via pointer event).

- `shortest` (default): same behavior as before, selects the shortest annotation at the click position.
- `all` new behavior: will select the whole stack of annotations at the click position, i.e. all annotations that intersect the position.

__In addition:__ this PR renames the `TextAnnotatorPopup` in the react package to `TextAnnotationPopup`. This makes naming consistent with Annotorious (and generally makes more sense I think). The old `TextAnnotatorPopup` component is still available in order to not break existing code. But is now marked as deprecated. Using it will log a runtime warning.


packages/text-annotator/src/state/TextAnnotationStore.ts
```diff
@@ -15,7 +15,9 @@ export interface TextAnnotationStore<T extends TextAnnotation = TextAnnotation>
 
   getAnnotationRects(id: string): DOMRect[];
 
-  getAt(x: number, y: number, filter?: Filter): T | undefined;
+  getAt(x: number, y: number, all: true, filter?: Filter): T[] | undefined;
+  getAt(x: number, y: number, all: false, filter?: Filter): T | undefined;
+  getAt(x: number, y: number, all?: boolean, filter?: Filter): T | T[] | undefined;
```
oleksandr-danylchenko, line 20: Should we use the `options: { all?: boolean, filter?: Filter }` type argument to avoid passing an unnamed boolean args like that: `store.getAt(event.clientX - x, event.clientY - y, false, currentFilter);`?

oleksandr-danylchenko, line 20: What is the reason behind using explicit `true` or `false` values for the `all` argument? 🤔

rsimon: TS will automatically infer the return type correctly. (As either an annotation or an array.)

packages/text-annotator/src/TextAnnotatorOptions.ts
```diff
@@ -17,6 +17,8 @@ export interface TextAnnotatorOptions<I extends TextAnnotation = TextAnnotation,
 
   presence?: PresencePainterOptions;
 
+  selectionMode?: 'shortest' | 'all';
```
oleksandr-danylchenko, line 20: Should we add documentation about meaning of those properties right in the code?
```suggestion
  /**
   * Defines the behavior when the user selects an existing annotation
   * that overlaps other annotations via the pointer event.
   * - `shortest` (default): selects the shortest annotation at the click position.
   * - `all`: will select the whole stack of annotations at the click position,
   *          in other words, all annotations that intersect the position.
   */
  selectionMode?: 'shortest' | 'all';
```

packages/text-annotator/src/state/TextAnnotationStore.ts
oleksandr-danylchenko: Ah, that makes sense 👍🏻
I didn't notice the return type difference between those signatures

Issue #179, opened 2024-11-11
`useSelection` hook acts differently in React 18 than in prior versions


oleksandr-danylchenko:
### TL;DR;
There's a difference in how the `useSelection` reports the `selection.clear()` in React 18 and the rest of the prior version. The cleanup isn't reported in the 18th version due to the [automatic set batching](https://github.com/reactwg/react-18/discussions/21). But in other versions, it's reported and can be breaking for the popup / garbage annotations collector implementation.
 
---

React 18 introduced a new automatic set state batching feature — https://github.com/reactwg/react-18/discussions/21. It allows deferring the rendering call until the function runs all the set methods within.

Its effect is noticeable in how the `useSelection` hook reports about the currently selected annotation:

### React 18

Here's the behavior when I select the text with the keyboard:
https://github.com/recogito/text-annotator-js/blob/2b00afd95d002f97f5421014a869ae226eb09194/packages/text-annotator/src/SelectionHandler.ts#L250-L252

https://github.com/user-attachments/assets/5d002f02-0f2a-422a-a5d2-996342466d4d

![image](https://github.com/user-attachments/assets/3d07d977-d8ce-44c8-9ca4-fc11a8c8e0ec)

### React 17, or less.

The same keyboard selection:

https://github.com/user-attachments/assets/cd2b8493-662a-48e3-b0fc-080e5f2f6161

![image](https://github.com/user-attachments/assets/9a920bf8-bbd3-4256-915a-fb1a977aa1c3)

---

**That's problematic for the React <18 apps** with a garbage collector component that automatically deletes annotations without a note or a highlight assigned. By listening to the `useSelection` that component will think that **_the selection was cleared and changed entirely_** when a user simply lifts the `Shift` key or releases the finger from the touchscreen handlebars.

## Suggested Changes
We can reduce the number of the `selection.clear()` calls and execute it only when:
- `selectionchange` event is emitted but there's no `currentTarget` present (already in place)
- On "Select All" handling
- On an arbitrary arrow press
- Click beyond the annotation range

**That will help to keep the same annotation in place during the continued selection with the keyboard or touch.**

PR #180, opened 2024-11-11
#179 Removed excessive selection cleanup

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/179

## Changes Made
I removed the excessive `selection.clean()` calls that can cause losing the same selected annotation reference. Previously upon releasing the `Shift` or a finger, the selection was cleared out and immediately updated. That cleanup would make the consumer think that a brand-new annotation was selected while the user just kept selecting the text after a pause.
Additionally, that cleanup is processed inconsistency in React 18 vs React <18 consumers.

Now, the selection is cleared only on:
- `selectionchange` event is emitted but there's no `currentTarget` present (already in place)
- On "Select All" handling
- On an arbitrary arrow press
- Click beyond the annotation range

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -219,7 +219,6 @@ export const SelectionHandler = (
         currentTarget = undefined;
         clickSelect();
       } else if (currentTarget && currentTarget.selector.length > 0) {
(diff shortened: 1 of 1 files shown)


rsimon:
Thanks!

PR #181, opened 2024-11-13
Generalized `W3CTextFormatAdapter`

oleksandr-danylchenko:
## Changes Made
I made the `W3CTextFormatAdapter` generalized for the package consumers.

For that, I followed the `TextAnnotator` and `createTextAnnotator` definition example:
https://github.com/recogito/text-annotator-js/blob/ff7100718a0ee9b0366942df0c65adba24646f42/packages/text-annotator/src/TextAnnotator.ts#L17
https://github.com/recogito/text-annotator-js/blob/ff7100718a0ee9b0366942df0c65adba24646f42/packages/text-annotator/src/TextAnnotator.ts#L30-L33

That will help to utilize custom W3C definitions on the consumer's side w/o triggering the following errors:
image
error details


rsimon:
Looks good, thanks!

PR #182, opened 2024-11-21
Reverted breaking API change, removal of the `denormalizeRectWithOffset`

oleksandr-danylchenko:
## Changes Made
Reverted the https://github.com/recogito/text-annotator-js/commit/945c09db538bf6bedfe4740be14bdeb872b275bf commit, where the `normalizeRects.ts` utility file was deleted.
However, its utilities are used by the consumers that implement custom popup components.


rsimon:
Sorry, in this case I have to contradict. I'm fine __if__ we find sensible names for it. But this way, when I looked at those methods for a second time, I had already forgotten what direction "normalize" vs. "denormalize" goes. Having two args of the same type doesn't help either. We're really talking about a single line of code here, and it's pretty obvious what it does if you see it. But with just generic "(de)normalize" as name, readability IMO suffers massively, for a single line of (readable) avoided code.

oleksandr-danylchenko:
Hmm, that's reasonable. Maybe I overprojected my familiarity with the "normalize/denormalize" functions.

Maybe something like "`toParentBounds`"/"`toViewportBound`" would be more appealing.

oleksandr-danylchenko:
cc @rsimon, could you please take a look at the PR in your spare time? 🙏🏻  
I don't wanna bring it into the fork too eagerly to prevent furthering the merge conflicts.

PR #184, opened 2024-12-06
Broaden the internal type generic usage

oleksandr-danylchenko:
## Issue
Back in https://github.com/recogito/text-annotator-js/commit/15186ec1841df071d4153c715ff8193341900600 @rsimon added support for the internal type generic. However, there are a few points where it was left unutilized: 
- `TextAnnotatorProps` for the `TextAnnotator` React component. The issue would show up if you try to pass the `Filter<TextAnnotation>` to the `filter` prop: image
- `W3CTextFormat`
- `createTextAnnotator`
- `HighlightStyleExpression`

### DIFF
--- a/packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts
+++ b/packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts
@@ -17,10 +17,10 @@ export type W3CTextFormatAdapter<I extends TextAnnotation = TextAnnotation, E ex
  * @param container - the HTML container of the annotated content,
  *                    Required to locate the content's `range` within the DOM
  */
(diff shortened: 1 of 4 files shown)


packages/text-annotator/src/model/w3c/W3CTextFormatAdapter.ts
```diff
@@ -171,6 +173,6 @@ export const serializeW3CTextAnnotation = <E extends W3CTextAnnotation = W3CText
     created: created?.toISOString(),
     modified: updated?.toISOString(),
     target: w3cTargets
-  } as E;
+  } as unknown as E;
```
oleksandr-danylchenko, line 176: The `as unknown` coercion is justified here to handle the `...rest` destruction above. It'll have the following broad type: 
image
So the TS won't be able to know that the `I` actually fulfills the `E` type constraint.

Issue #185, opened 2024-12-16
Popup Currently Not Working on Android


rsimon:
Hm... looks like the popup from the `@recogito/react-text-annotator` package is currently not working on Android. (Seems to work ok on desktop + iOS.)

oleksandr-danylchenko:
Yeah, I can reproduce it locally 👀

oleksandr-danylchenko:
Yeah, on creation the `bounds` are undefined for the annotation 🤔 
![image](https://github.com/user-attachments/assets/a801107a-cff6-42c4-9538-498999d8cf95)
https://github.com/recogito/text-annotator-js/blob/5b9abad31724db14d473f7d363b295a967cd93b7/packages/text-annotator-react/src/TextAnnotationPopup/TextAnnotationPopup.tsx#L95-L98

oleksandr-danylchenko:
I think it went unnoticed in my app, as I also added the "idling" check before showing the popup - https://github.com/recogito/text-annotator-js/pull/174. So it gives enough time for the `bounds` to get populated before we query them.

PR #186, opened 2024-12-16
#185 Fixed popup rendering on Android

oleksandr-danylchenko:
## Issues
See - https://github.com/recogito/text-annotator-js/issues/185#issuecomment-2545199474

In the https://github.com/recogito/text-annotator-js/pull/123 I made the popup presence check resilient to the underlying collapsed ranges in the target's selectors.

However, it caused a timing regression on Android. When the annotation is processed in the popup's `useEffect`, the `target` and its bounds aren't yet populated.
https://github.com/recogito/text-annotator-js/blob/5b9abad31724db14d473f7d363b295a967cd93b7/packages/text-annotator-react/src/TextAnnotationPopup/TextAnnotationPopup.tsx#L95-L98
image

Also, the timing of the selection change event that populates the `target` differs on Android:
| Android | Mac |
|--------|--------|
| image | image | 

## Changes Made
I added a check for the selectors' presence on the annotation's target. That way, the effect will react properly to their addition and re-run the bounds querying.

### DIFF
--- a/packages/text-annotator-react/src/TextAnnotationPopup/TextAnnotationPopup.tsx
+++ b/packages/text-annotator-react/src/TextAnnotationPopup/TextAnnotationPopup.tsx
@@ -93,13 +93,18 @@ export const TextAnnotationPopup = (props: TextAnnotationPopupProps) => {
   const { getFloatingProps } = useInteractions([dismiss, role]);
 
   useEffect(() => {
-    if (annotation?.id) {
+    if (!r) return;
+
+    const annotationId = annotation?.id;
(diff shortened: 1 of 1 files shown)


rsimon:
I need to look at this tomorrow with a fresh mind (it's been a long day). But whatever you did - it's working, thanks! :-)

oleksandr-danylchenko:
Could this change be merged? I believe that Android users still experience the missing popup when the `TextAnnotationPopup.tsx` is used.

rsimon:
Yikes, forgot about that. Looks good, thanks!

PR #187, opened 2024-12-16
Added missing `adapter` forwarding to the `createSelectionState`

oleksandr-danylchenko:
## Issue
The `createTextAnnotatorState` wasn't forwarding the provided `adapter` to the `createSelectionState`. That led to the missing serialization on the [`onUserSelect`](https://github.com/annotorious/annotorious/blob/94624c0dea045db435d30ba054a29e5c97f83d22/packages/annotorious-core/src/state/Selection.ts#L157-L164) call.

### Example
image
However, an internal type is provided to the `userSelectActionCallback`:
image

## Changes Made
Added `opts.adapter` forwarding to the `createSelectionState`. That makes the code compatible with the API changes made in https://github.com/annotorious/annotorious/commit/79c7b17f2002ff2dfcc93880c9048c8a73ad01e2.
I borrowed that approach from the `a9s/a9s`'s `createImageAnnotator`, which passes the whole `opts` object to the `createImageAnnotatorState`, [source](https://github.com/annotorious/annotorious/blob/94624c0dea045db435d30ba054a29e5c97f83d22/packages/annotorious/src/Annotorious.ts#L63).

### DIFF
--- a/packages/text-annotator/src/TextAnnotator.ts
+++ b/packages/text-annotator/src/TextAnnotator.ts
@@ -42,8 +42,7 @@ export const createTextAnnotator = <I extends TextAnnotation = TextAnnotation, E
     user: createAnonymousGuest()
   });
 
-  const state: TextAnnotatorState<I, E> =
-    createTextAnnotatorState<I, E>(container, opts.userSelectAction);
+  const state: TextAnnotatorState<I, E> = createTextAnnotatorState<I, E>(container, opts);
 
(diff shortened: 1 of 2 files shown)


oleksandr-danylchenko:
@rsimon, unfortunately, it's still the issue on the `main` branch. And it can be quite error-prone for any consumers that depend on the `onUserSelect`.

rsimon:
Looks good, thanks!

PR #190, opened 2025-02-18
Re-added updated `mergeClientRects`

oleksandr-danylchenko:


rsimon:
I think this one's no longer needed now?

Issue #191, opened 2025-02-20
The bundled `dist` directory includes the 3rd-party packages


oleksandr-danylchenko:
## Issue
⚠ When the `text-annotator-react` is being bundled, it doesn't exclude the external dependencies from the bundle ⚠

| `dist` | Extenal dependency bundled |
|--------|--------|
| ![i](https://github.com/user-attachments/assets/4638a9d3-579a-4413-bb90-e24e3d2daa6d) | ![i2](https://github.com/user-attachments/assets/7f84fc1e-28fd-42f0-b1d6-018d57934e79) | 

That needlessly bloats the bundle and doesn't utilize the dependencies installed by the consumer.

## Suggested Changes
I faced a similar challenge in my lib too. Unfortunately, the default `rollupOptions.external` doesn't help much, as they exclude only the peer deps, and listing all the newly added deps manually is tedious. Therefore, I used the [vite-plugin-externalize-deps](https://github.com/davidmyersdev/vite-plugin-externalize-deps)

PR #192, opened 2025-02-20
Externalized 3rd-party dependencies

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/191
> ⚠ When the text-annotator-react is being bundled, it doesn't exclude the external dependencies from the bundle ⚠

## Changes Made
Added `vite-plugin-externalize-deps` plugin that helps to prevent bundling 3rd-party dependencies. Instead, the consumer should provide them for the packages, so the latter can simply import them.

## Comparison
- `@recogito/text-annotator`

| Before | After |
|--------|--------|
| image | image | 

- `@recogito/react-text-annotator`

| Before | After |
|--------|--------|
| image | image |


packages/text-annotator-react/vite.config.ts
```diff
@@ -2,42 +2,33 @@ import { defineConfig } from 'vite';
...
     rollupOptions: {
-      external: [
-        ...Object.keys(packageJson.peerDependencies),
-        "@annotorious/core",
-        "@annotorious/react",
-        "@recogito/text-annotator",
-        "@recogito/text-annotator-tei"
-      ],
       output: {
-        preserveModules: true,
```
oleksandr-danylchenko, line 37: I believe that this property is redundant here. Unfortunately, the `preserveModules: true` doesn't preserve the original file structure of the package in the output, but simply splits it into quite a few chunks. To actually preserve it a more complex config is required, for example:
```ts
...
rollupOptions: {
	/**
	 * If you want to convert a set of files to another format while maintaining the file structure
	 * and export signatures, the recommended way—instead of using output.preserveModules that
	 * may tree-shake exports as well as emit virtual files created by plugins—is to
	 * turn every file into an entry point.
	 * @see https://rollupjs.org/configuration-options/#input
	 */
	input: Object.fromEntries(
		globSync('src/**/*.{ts,tsx}', {
			ignore: ['src/**/*.d.ts', 'src/**/*.stories.tsx', 'src/storybook/**/*']
		}).map((file) => [
			// 1. The name of the entry point
			// lib/nested/foo.js becomes nested/foo
			relative('src', file.slice(0, file.length - extname(file).length)),
			// 2. The absolute path to the entry file
			// lib/nested/foo.ts becomes /project/lib/nested/foo.ts
			fileURLToPath(new URL(file, import.meta.url))
		])
	),
	...
```
However, I believe it's an overkill for the package, as it still has only a single entry point. Also having the sourcemaps will help the browser to navigate through the original structure.

packages/text-annotator/vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+      output: {
+        assetFileNames: 'text-annotator.[ext]',
+        globals: {
+          'colord': 'Colord',
+          'uuid': 'UUID',
+          'dequal/lite': 'DequalLite',
+          '@annotorious/core': 'AnnotoriousCore',
+          'rbush': "RBush",
+          'hotkeys-js': 'HotkeysJs',
+        }
```
oleksandr-danylchenko, line 32: The globals are needed to bundle the UMD package using predictable names for the external packages. Otherwise, they will be guessed by Vite:
image

oleksandr-danylchenko:
I also verified that it doesn't break the existing consumers on our side ✅

rsimon:
Thanks, I'll give that a try ;-)

PR #195, opened 2025-04-04
Fixed unexpected deletion of the default `r6o-annotatable` & `no-focus-outline` classes on the React `container`

oleksandr-danylchenko:
## Issue
When a consumer app uses a CSS-in-JS solution, a different `className` can be passed to the `TextAnnotator` at runtime. Unfortunately, it'll cause the deletion of the default `r6o-annotatable` class that's added only once during the renderer initialization. That way, styling for the highlights layer will be lost ⚠️
| Right after initialization | After conditionally passed `className` |
|--------|--------|
| CleanShot 2025-04-04 at 14 30 55 | CleanShot 2025-04-04 at 14 29 49 | 

## Changes Made
Cemented in the default classnames on the React annotatable container. So any custom classes will be appended to the existing ones instead of entirely overriding them.


rsimon:
Good catch - thanks!

PR #197, opened 2025-04-15
Added support for the custom `scrollParent` for the `scrollIntoView`

oleksandr-danylchenko:
## Issue
The `scrollIntoView` is great and convenient in most cases, especially if you need to scroll to the annotation on load of the page. Thanks!
However, I our mobile app that uses `Ionic` components library, we faced an edge case with the `getScrollParent`. The `Ionic` handles a scrollable content in a bit of a peculiar way. It uses `#shadow-dom` to hide the "`inner-scroll`" scrollable element, but its "public" elements all have the `overflow: visible`[^1]
CleanShot 2025-04-15 at 12 36 13

Therefore, if we render the `text-annotator` over such content, the `scrollIntoView` will be ineffective, operating over non-scrollable elements 🤷🏻‍♂️ 

## Changes Made
Added support for providing a custom `scrollParent` to overcome the sensible limitations of the `getScrollParent`.

[^1]: https://ionicframework.com/docs/api/content#scroll-methods

### DIFF
--- a/packages/text-annotator/src/api/scrollIntoView.ts
+++ b/packages/text-annotator/src/api/scrollIntoView.ts
@@ -1,9 +1,9 @@
 import type { TextAnnotationStore } from '../state';
-import type { TextAnnotation, TextAnnotationTarget } from '../model';
+import type { TextAnnotation } from '../model';
 import { reviveTarget } from '../utils';
 
 const getScrollParent = (el: Element) => {
-  if (el === null)
+  if (!el)
     return document.scrollingElement;
 
   const { overflowY } = window.getComputedStyle(el);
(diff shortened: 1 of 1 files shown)


packages/text-annotator/src/api/scrollIntoView.ts
```diff
@@ -16,59 +16,68 @@ const getScrollParent = (el: Element) => {
...
-    const parentWidth = scrollParent.clientWidth;
+  /*
+   Note: `getBoundingClientRect` seems to return the wrong height!
+   (Includes block elements?)
+   Instead, the normalized height is used from the spatial index!
+  */
+  const { width, height } = store.getAnnotationBounds(target.annotation);
 
-    // Position of the annotation relative to viewport
-    // Note: first selector is not necessarily top one...
```
oleksandr-danylchenko, line 33: This line seems contradictory to the following one 😅 https://github.com/recogito/text-annotator-js/blob/c07bdaf5a91b3e804c8fb2f4bb0f07b8481f6fd1/packages/text-annotator/src/api/scrollIntoView.ts#L61-L62

oleksandr-danylchenko:
Also, due to such `scroll` handling, I needed to add the following `redraw` workaround:
```tsx
const AnnotatorScrollRedraw: FC<Pick<Props, 'scrollElement'>> = (props) => {
	const { scrollElement } = props;

	const r = useAnnotator();

	useEffect(() => {
		if (!scrollElement || !r) return;

		const handleScroll = () => r.redraw(true);
		scrollElement.addEventListener('scroll', handleScroll, { capture: true, passive: true });

		return () => scrollElement.removeEventListener('scroll', handleScroll);
	}, [scrollElement, r]);

	return null;
};
```

(See https://github.com/recogito/text-annotator-js/pull/104)

PR #198, opened 2025-05-15
Added non-breaking handling for the missing  `annotation` or `scrollParent` for the `scrollIntoView`

oleksandr-danylchenko:
## Issue
When a user passes an ID of an annotation that's not present in the store to the `scrollIntoView`, a hard exception will be raised. If it's left unhandled, it can cause the whole app to be replaced with the WSOD. In my view, it would be more sensible to make the error handling more mild in this case. The scroller can simply warn a user that it doesn't where to scroll.

###### (Soomo Staged)


rsimon:
Thanks, makes sense! I think we've actually encountered the same thing (when users share links that include an annotation ID for selection, but that annotation had meanwhile been deleted.)

oleksandr-danylchenko:
> when users share links that include an annotation ID for selection, but that annotation had meanwhile been deleted

That's almost exactly the case we faced too 🙌🏻

Issue #202, opened 2025-06-11
The hover-based style isn't applied on hover


oleksandr-danylchenko:
Hovering over an annotation doesn't make it redraw! Which makes the following `style` do nothing when a user hovers over an annotation:
```ts
const style = ((annotation, state, z) =>
  ({
    fillOpacity: state.hovered ? 0.9 : 0.2,
    ...
  }));
```
![CleanShot 2025-06-11 at 18 54 12](https://github.com/user-attachments/assets/39b88620-1db8-4daa-a5bc-a2306faa3c1e)

_Originally posted by @oleksandr-danylchenko in https://github.com/recogito/text-annotator-js/issues/201#issuecomment-2963363309_

PR #203, opened 2025-06-13
#202 Added `hover` state subscription for the `redraw`

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/202

## Changes Made
Following the `selection` subscription example
https://github.com/recogito/text-annotator-js/blob/8f46806defde7b6ef93c0d9811b4fd5c9afc9100/packages/text-annotator/src/highlight/baseRenderer.ts#L131-L132
I added `redraw` calls on the `hovered` state change. Now, the highlights can be properly styled when they are hovered:
CleanShot 2025-06-13 at 13 46 38


rsimon:
Thanks!

PR #205, opened 2025-08-25
#133 click outside container after selection

SovanramyVar4D:
Should fix https://github.com/recogito/text-annotator-js/issues/133
- Fix click outside container triggering createAnnotation event
- Add delete annotation button on test page

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -14,7 +14,8 @@ import {
   isMac,
   isWhitespaceOrEmpty,
   trimRangeToContainer,
-  isNotAnnotatable
+  isNotAnnotatable,
+  isOutsideContainer
 } from './utils';
 
 const CLICK_TIMEOUT = 300;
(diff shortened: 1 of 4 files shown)


rsimon:
Thanks! I'm wondering though if it makes sense to simply combine this check with the normal `isNotAnnotatable` function. If the event is outside the container, it should be considered "not annotatable" after all?

SovanramyVar4D:
> Thanks! I'm wondering though if it makes sense to simply combine this check with the normal `isNotAnnotatable` function. If the event is outside the container, it should be considered "not annotatable" after all?

Yes it was also an option I considered, but I think it should be 2 distinct use cases but often closely related.

SovanramyVar4D:
I don't really have any concrete example, it just "might" exist
but you know your library better than me, so I trust your judgement

rsimon:
Hi @SovanramyVar4D, I just published [an alternative PR](https://github.com/recogito/text-annotator-js/pull/208) to address issue #133 - let me know what you think.

rsimon:
Closing this in favour of #208

Issue #206, opened 2025-08-27
Discontinuous Select/Annotate


rsimon:
New requirement: the user wants to select a discontinuous piece of text: 

```
Lorem [ipsum dolor] sit amet, [consectetur] adipiscing elit. 
       -----------             -----------
```

In this case, the two highlights **together** are considered **a single annotation**. (One target with two selectors in our internal model.)

From a UX perspective, my feeling is we'd simply allow `CTRL`/`Cmd` + select. If an annotation is already selected, we'd modify it instead of creating a new one.

Also CC'ing @oleksandr-danylchenko: is this something you already have as a use case in your application, perhaps?

rsimon:
Bonus challenge: Chrome does not actually support native multi-range select via `Cmd` + select! (Firefox does.)

rsimon:
A crude proof-of-concept is available [in this draft PR](https://github.com/recogito/text-annotator-js/pull/207).

**Example:**
- Create an annotation.
- While the annotation is selected, hold `Ctrl` (or `Cmd` on Mac) and make another selection.
- This selection will get **added to the selected annotation** rather than create a new annotation.
- This works both for click-and-drag as well as double-click (to select a single word)

https://github.com/user-attachments/assets/48738faa-82e4-4037-b3c5-adcf94858d52

oleksandr-danylchenko:
> Also CC'ing @oleksandr-danylchenko: is this something you already have as a use case in your application, perhaps?

Unfortunately, no. In my app, we don't support discontinuous annotating. Inversely, we rely on its continuity to properly migrate them when the annotated content changes.

PR #207, opened 2025-08-27
Discontinuous select

rsimon:
This is a (very) crude prototype for multi-selection, i.e. creating a single annotation from multiple disjoint parts. Works on (desktop) Chrome and Firefox. See issue #206 for more information.

**DRAFT** just for comparison with the main branch so far

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -46,7 +46,10 @@ export const SelectionHandler = (
 
   const { store, selection } = state;
 
-  let currentTarget: TextAnnotationTarget | undefined;
+  let currentTarget: {
+    previous?: TextAnnotationTarget,
+    next: TextAnnotationTarget
+  } | undefined;
 
   let isLeftClick: boolean | undefined;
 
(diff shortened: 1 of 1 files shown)


packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -46,12 +46,20 @@ export const SelectionHandler = (
 
   const { store, selection } = state;
 
-  let currentTarget: TextAnnotationTarget | undefined;
+  let currentTarget: {
+    previous?: TextAnnotationTarget,
```
rsimon, line 50: Not ideal (and horrible naming). But we need to record the state of the annotation *before* this segment was added.

packages/text-annotator-react/src/TextAnnotationPopup/TextAnnotationPopup.tsx
```diff
@@ -87,12 +93,6 @@ export const TextAnnotationPopup = (props: TextAnnotationPopupProps) => {
     whileElementsMounted: autoUpdate
   });
 
-  const dismiss = useDismiss(context);
-
-  const role = useRole(context, { role: 'dialog' });
-
-  const { getFloatingProps } = useInteractions([dismiss, role]);
```
rsimon, line 94: @oleksandr-danylchenko I noticed that `getFloatingProps` isn't actually used. Is this an omission? Or is this no longer needed now we're using the FloatingPortal & FloatingFocusManager?

PR #217, opened 2025-09-22
#133 Fixed dismissed selection when pointer ends on a gutter

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/133#issuecomment-3318287874

> Unfortunately, I noticed a _UX regression_ 👀
> 
> Now, a user can't make an annotation if a pointer ends up on a page's gutter. Instead, for the selection to be effective, a user needs to keep the selection entirely within an annotatable container.
> 
> https://github.com/user-attachments/assets/221adeec-98b2-4651-8e16-a3394457d415
> 
> That's because, since the https://github.com/recogito/text-annotator-js/pull/208, everything outside of the `container` is considered as not-annotatable. That causes the following condition to trigger and dismiss the selection that ends on a gutter:
> https://github.com/recogito/text-annotator-js/blob/574bc283581b90d8da7de071bad415c26ca38338/packages/text-annotator/src/SelectionHandler.ts#L197-L205

---

I realized that we intend to dismiss the selection only when a user _clicks_ on a not-annotatable. That was also the intention of the `dismissOnClick` in the https://github.com/recogito/text-annotator-js/pull/208. 

However, by listening ONLY to the `pointerup` event and then checking the `target` doesn't allow us to differentiate between an active selection and a click. That way, the ability to annotate when a user finishes their selection on a gutter _was accidentally removed_.

## Changes Made
I added the `pointerdown` listener to the `document` level, as it was initially done in the https://github.com/recogito/text-annotator-js/pull/135. That way, we can handle the "click" event anywhere on the screen.
Then, we can check whether _a user's click_ lands on a not-annotatable element and whether we should prevent any further processing. 

Otherwise, we'll either have a broken selection handling on a gutter, or we won't be able to dismiss a selection by clicking on a gutter - https://github.com/recogito/text-annotator-js/pull/217#issuecomment-3318450281.

## Demo

https://github.com/user-attachments/assets/d0af453b-b100-4909-bb79-84865104ef86

###### (Soomo Staged 22.09.25)

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -192,13 +192,15 @@ export const SelectionHandler = (
   const onPointerUp = async (evt: PointerEvent) => {
     if (!isLeftClick) return;
 
+    const lastUpEvent = clonePointerEvent(evt);
+
     // Logic for selecting an existing annotation
     const clickSelect = () => {
       const { x, y } = container.getBoundingClientRect();
 
-      if (isNotAnnotatable(container, evt.target as Node)) {
+      if (isNotAnnotatable(container, lastUpEvent.target as Node)) {
         const shouldDismissSelection = typeof dismissOnNotAnnotatable === 'function'
-          ? dismissOnNotAnnotatable(evt, container)
+          ? dismissOnNotAnnotatable(lastUpEvent, container)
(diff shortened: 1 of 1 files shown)


rsimon:
Nice catch - thanks!

Issue #221, opened 2025-09-25
The selection that starts from a gutter doesn't create an annotation consistently


oleksandr-danylchenko:
I think that there's another UX quirk that appeared due to handling everything outside as not-annotatable. When a user starts their selection from the page's gutter, the annotation won't be preserved due to the same `isNotAnnotatable(container, evt.target as Node)` check:

https://github.com/user-attachments/assets/7f8ec9c0-d26b-4888-9d73-2be2b80baccd

---

UPD: It's rather weirdly behaving, though. When I dismiss the selection on the annotatable container, the following selections will be preserved when I make them rather rapidly:

https://github.com/user-attachments/assets/12dbbb74-a5db-4459-89c2-894d31693eec

_Originally posted by @oleksandr-danylchenko in https://github.com/recogito/text-annotator-js/issues/150#issuecomment-3333080513_

Issue #220, opened 2025-09-25
The "select all" keyboard action doesn't work anymore


oleksandr-danylchenko:
@rsimon, the general selection behavior has been improved since we started treating all the elements beyond the annotatable container as `not-annotatable` (https://github.com/recogito/text-annotator-js/pull/208). Therefore, now the selection outside won't cause accidental coverage of the whole container.

---

Although now it seems like we lost the ability to select the text with the "select all" command 👀 

<img width="1050" height="316" alt="CleanShot 2025-09-25 at 12 11 37@2x" src="https://github.com/user-attachments/assets/bff2793e-443e-44cf-bd9a-f3d00801efb5" />
<img width="1651" height="266" alt="CleanShot 2025-09-25 at 12 11 47" src="https://github.com/user-attachments/assets/f9a9c16c-7a3d-4e6f-803c-f6a0daa04362" />

That's because the selection usually starts outside of the container, which makes the following condition assign `undefined` to the `currentTarget`: https://github.com/recogito/text-annotator-js/blob/34c345afaaba00ca7bcb57be33ce126a3a92597a/packages/text-annotator/src/SelectionHandler.ts#L65-L78

_Originally posted by @oleksandr-danylchenko in https://github.com/recogito/text-annotator-js/issues/150#issuecomment-3333023223_

PR #224, opened 2025-10-01
Selection Behavior

rsimon:
## In this PR

This (draft) PR is for discussing problems related to PR #150 and issues #220 and #221.

**Key problem:** no annotation is created if the selection starts **before** the annotatable content. This causes both issue #221 and (in general) causes CTRL+A to not work. (FWIW: @oleksandr-danylchenko I believe the reason you are sometimes getting annotations to work in #221 is because you're moving so quickly that the pointerdown event ends up being on the annotatable content.)

The reason annotations get discarded when the selection starts outside is here:

https://github.com/recogito/text-annotator-js/blob/653113495ae9f70b2688793748139cf8d9ffdc72/packages/text-annotator/src/SelectionHandler.ts#L97-L105

**First,** we can change this, and check whether `sel.anchorNode` **AND** `sel.focusNode` is not annotatable. This is an improvement. But not a 100% guarantee that a selection that includes annotatable content will pass through this condition. Alternatively, we can check for an actual intersection between selection and container element. (Does this have any unwanted side effects?)

**Next**, there is this condition:

https://github.com/recogito/text-annotator-js/blob/653113495ae9f70b2688793748139cf8d9ffdc72/packages/text-annotator/src/SelectionHandler.ts#L71-L78

This will prevent a `currentTarget` from getting started if the event isn't on annotatable content. I think we *may* actually be able to simply drop this requirement, given the improvements we've recently made to selection handling. But we definitely have to check for regressions and unintended side effects.

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -58,24 +58,16 @@ export const SelectionHandler = (
 
   let lastDownEvent: Selection['event'] | undefined;
 
-  const onSelectStart = (evt: Event) => {
+  const onSelectStart = () => {
     if (isLeftClick === false)
       return;
 
-    /**
-     * Make sure we don't listen to selection changes that were
-     * not started on the container, or which are not supposed to
-     * be annotatable (like a component popup).
-     * Note that Chrome/iOS will sometimes return the root doc as target!
-     */
-    currentTarget = /*isNotAnnotatable(container, evt.target as Node)
-      ? undefined
-      :*/ {
-        annotation: uuidv4(),
-        selector: [],
-        creator: currentUser,
-        created: new Date()
-      };
+    currentTarget = {
+      annotation: uuidv4(),
+      selector: [],
+      creator: currentUser,
+      created: new Date()
+    };
   };
 
   const onSelectionChange = debounce((evt: Event) => {
@@ -117,11 +109,11 @@ export const SelectionHandler = (
     if (lastDownEvent?.type === 'pointerdown') {
       if (timeDifference < 1000 && !currentTarget) {
         // Chrome/iOS does not reliably fire the 'selectstart' event!
-        onSelectStart(lastDownEvent || evt);
+        onSelectStart();
       } else if (sel.isCollapsed && timeDifference < CLICK_TIMEOUT) {
         // Firefox doesn't fire the 'selectstart' when user clicks
         // over the text, which collapses the selection
-        onSelectStart(lastDownEvent || evt);
+        onSelectStart();
       }
     }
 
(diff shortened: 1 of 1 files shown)


packages/text-annotator/src/utils/trimRangeToContainer.ts
```diff
@@ -29,35 +29,38 @@ export const trimRangeToContainer = (
...
     }
   }
+  */
 
   /**
    * If the range starts outside the container -
    * trim it to the start of the container
    */
-  if (!containsRangeStart) {
+  if (!container.contains(trimmedRange.startContainer)) {
```
rsimon, line 55: I may be wrong. But I still think the change from PR #150 isn't actually needed?

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -99,7 +99,8 @@ export const SelectionHandler = (
      * another element in a not-annotatable area. A rare case in practice.
      * But rich text editors like Quill will do it!
      */
-    if (isNotAnnotatable(container, sel.anchorNode)) {
+    // if (isNotAnnotatable(container, sel.anchorNode) && isNotAnnotatable(container, sel.focusNode)) {
+    if (!sel.getRangeAt(0).intersectsNode(container)) {
```
rsimon, line 103: This check might make more sense than checking the start/end nodes.

oleksandr-danylchenko:
Thanks for the PR, @rsimon 🙌🏻 
Today I'm having a day off, away from the desktop. So I'll be able to try out the changes and get back with feedback tomorrow.

packages/text-annotator/src/SelectionHandler.ts
```diff
@@ -94,12 +94,16 @@ export const SelectionHandler = (
...
+      Array.from(Array(sel.rangeCount).keys()).map(idx => sel.getRangeAt(idx));
+
     /**
      * This is to handle cases where the selection is "hijacked" by
      * another element in a not-annotatable area. A rare case in practice.
      * But rich text editors like Quill will do it!
      */
-    if (isNotAnnotatable(container, sel.anchorNode)) {
+    // if (isNotAnnotatable(container, sel.anchorNode) && isNotAnnotatable(container, sel.focusNode)) {
+    if (!selectionRanges.some(r => r.intersectsNode(container))) {
```
rsimon, line 106: This check might make more sense than checking the start/end nodes.

packages/text-annotator/src/utils/trimRangeToContainer.ts
oleksandr-danylchenko: Considering the addition of the following check I agree that we no longer need the change from the #150 ✅ 
https://github.com/recogito/text-annotator-js/blob/f80a78306ddad16d112ba76299af9ca366c1a192/packages/text-annotator/src/SelectionHandler.ts#L106

oleksandr-danylchenko (approved):
Thanks, @rsimon! The changes look sensible. I llike that now the we check for the partial overlap with the annotatable container instead of relying on the `range.anchorNode`. That should fix the selection handling when it starts/ends on a not-annotatable elements outside of the container:

https://github.com/user-attachments/assets/0f56434c-4cee-4695-a337-faa4e88715b7

PR #227, opened 2025-10-06
#226 Added empty text node content check before processing

### LINKED ISSUE #226: The new `getHighlightClientRects` skips the selection ranges that cover a single element
> Unfortunately, this method seems to contain a regression that causes the selection in my app to skip paragraphs in the middle of it:
> 
> https://github.com/user-attachments/assets/30e37cbb-dc66-4182-93d9-3450589d138b

---

In my example, each element with the `data-webtexts-readable-id` creates a separate range:

Image

Then the `spatialTree` goes over them in the `toItems` method:
https://github.com/recogito/text-annotator-js/blob/1c34de0d879e7c2d8379d0305c8c48d1bde2c468/packages/text-annotator/src/state/spatialTree.ts#L42-L45

Unfortunately, I noticed that if a range covers an element with a single `text` node, that range will be effectively discarded in the [`getHighlightClientRects`](https://github.com/recogito/text-annotator-js/blob/1c34de0d879e7c2d8379d0305c8c48d1bde2c468/packages/text-annotator/src/utils/getHighlightClientRects.ts#L1). That's because, for some reason, such a range will have TWO matching `text` nodes instead of one. However, only the first one is actually reachable in the DOM. That will trigger `textNodes.length >= 2` logic branch, which will create a collapsed `firstRange`:

Image

As a result, no client rects will be collected for the `range` 🤷🏻‍♂️.

 _Originally posted by @oleksandr-danylchenko in [c2e7731](https://github.com/recogito/text-annotator-js/commit/c2e7731840dc3408a0a424f21b2764c2dff2fd7d#r167237434)_

oleksandr-danylchenko:
## [Issue](https://github.com/recogito/text-annotator-js/issues/226)
Fixes https://github.com/recogito/text-annotator-js/issues/226

In some of the consumer apps, the DOM layout may contain collapsed `\n` symbols that are not clearly visually represented in the dev tools. However, they are considered in the following interator:
```ts
document.createNodeIterator(range.commonAncestorContainer, NodeFilter.SHOW_TEXT)
```
image

Those empty nodes inclusion makes the `getHighlightClientRects` to incorrectly process a trivial case, as multi-node selection. Which, in turn, causes skipping selected elements from highlighting.

See - https://github.com/recogito/text-annotator-js/issues/226#issuecomment-3371007681

## Changes Made
Excluded empty text nodes from highlights collection.

###### (Soomo staged, 06.10.25)

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -13,7 +13,7 @@ import {
   splitAnnotatableRanges,
   rangeToSelector,
   isMac,
-  isWhitespaceOrEmpty,
+  isRangeWhitespaceOrEmpty,
   trimRangeToContainer,
   isNotAnnotatable
 } from './utils';
@@ -150,7 +150,7 @@ export const SelectionHandler = (
       selectionRanges.map(r => trimRangeToContainer(r, container));
 
     // The selection should be captured only within the annotatable container
-    if (containedRanges.every(r => isWhitespaceOrEmpty(r))) return;
+    if (containedRanges.every(r => isRangeWhitespaceOrEmpty(r))) return;
 
     const annotatableRanges = containedRanges.flatMap(r => splitAnnotatableRanges(container, r.cloneRange()));
 
--- a/packages/text-annotator/src/utils/getHighlightClientRects.ts
+++ b/packages/text-annotator/src/utils/getHighlightClientRects.ts
@@ -1,23 +1,29 @@
+import { isNodeWhitespaceOrEmpty } from './isWhitespaceOrEmpty';
+
 export const getHighlightClientRects = (range: Range) => {
   const textNodes: Text[] = [];
 
   // Get all text nodes inside the range's commonAncestorContainer
   const it = document.createNodeIterator(range.commonAncestorContainer, NodeFilter.SHOW_TEXT);
 
-  // Filter text nodes that intersect the range. Note that we could
-  // also include this filter into the node iterator directly. But
-  // The while loop is faster (!) - possibly due to function call overhead.
+  /*
+   Filter text nodes that intersect the range. Note that we could
+   also include this filter in the node iterator directly.
+   But the while loop is faster (!) - possibly due to a function call overhead.
+  */
   let currentNode: Text | undefined;
  
   while ((currentNode = it.nextNode() as Text)) {
-    if (range.intersectsNode(currentNode))
+    if (range.intersectsNode(currentNode) && !isNodeWhitespaceOrEmpty(currentNode)) {
       textNodes.push(currentNode);
+    }
   }
 
   if (textNodes.length < 2) {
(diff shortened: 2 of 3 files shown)


oleksandr-danylchenko:
> Yikes - this one has some crazy regressions. I'm investigating...

Wow 👀 
It passed the QC on our side, btw

rsimon:
**v3.3.1** (how it should be)

Bildschirmfoto 2025-10-07 um 14 31 49

**v3.3.2**

Bildschirmfoto 2025-10-07 um 14 30 05

rsimon:
Not sure how to best address this... @oleksandr-danylchenko do you have a simple reproducible example for your case? 

I guess we could track if the first/last element got dropped or (better) filter in the adjustment stage?

oleksandr-danylchenko:
> Was this change meant as an internal optimization only?

No, not really. The highlighting functionality was broken when a range covered a single `p` element that should have contained a single `text` node, respectively. However, in my app, the selection also captures another `text` node that contains the `\n` symbol.
image
Unfortunately, it makes the range adjustment skip the actual node with a text, creating a collapsed `firstRange` - https://github.com/recogito/text-annotator-js/issues/226#issuecomment-3371007681

oleksandr-danylchenko:
> @oleksandr-danylchenko do you have a simple reproducible example for your case?

Unfortunately, nope. Not at the moment, at least. But I can try to assemble the one once I'll understand what causes the addition of those `\n` symbols.

rsimon:
Hm, not sure I understand how that can happens. But either way: capturing an empty text node shouldn't be the problem, and I don't think we should artificially remove them. The question, really, is how to reliably adjust the range.

oleksandr-danylchenko:
@rsimon, I was able to replicate the issue in the `text-annotator-react` by adding the following lines for the `/test/App.tsx`:
```tsx
<p>First paragraph</p>
{"\n"}
<span className="not-annotatable"/>
<p>Second paragraph</p>
{"\n"}
<span className="not-annotatable"/>
<p>Third paragraph</p>
{"\n"}
<span className="not-annotatable"/>
```
Then, I can get the same "empty" `\n` text node captured:
CleanShot 2025-10-07 at 17 45 47

rsimon:
Actually if you use this markup, it's the same problem. Remove the `not-annotatable` and it works. (At least that's what I'm seeing.)


```
<p>First paragraph</p>
<span className="not-annotatable" />
<p>Second paragraph</p>
<p>Third paragraph</p>
```

rsimon:
This definitely seems to point to an issue in the range adjustment logic. The `not-annotatable` will split the annotation into two selectors, which will then get processed separately. And then the range start adjustment drops the first text node for some reason. In a way, it must be the same thing (or something very close to) what I've been seeing after applying the whitespace filter. Need to log off now, but now that we nailed down a reproducible case, I can look into it tomorrow morning!

PR #240, opened 2025-11-26
Fixed programmatically focusable `tabindex` assignment

oleksandr-danylchenko:
## Issue
I noticed a mistake in the `tabindex` assignment logic. The attribute was assigned if an element didn't have it _AND the `tabIndex` is less than 0_. The latter can't be judged reliably if there's no `tabindex` on an element 🤷🏻‍♂️


rsimon:
Thx - I also added the change to the v4 branch!

PR #242, opened 2025-11-26
Added `userSelectAction` forwarding through the `TextAnnotator` component

oleksandr-danylchenko:
## Issue
The `anno?.setUserSelectAction` was originally added in the https://github.com/recogito/text-annotator-js/pull/126. Yet it was accidentally removed during the conflict resolution in https://github.com/recogito/text-annotator-js/pull/126/commits/1e5ada494085f701d6adbba38005e5c9715269d9

Instead, the `userSelectAction` should be forwarded to the annotated instance through the `TextAnnotator` component, as the rest of the interactive props.


rsimon:
Yikes, sorry for this. Also merged with the v4 branch!

PR #243, opened 2026-01-04
Remove UA parser for simpler navigator.platform check

jonathanKingston:
ua-parser-js uses AGPL-3 license and it seems like there's no need for such a heavy dependency to check for Mac support. 

Adding this dependency essentially makes the whole library AGPL-3 which I'm assuming is not intentional.


oleksandr-danylchenko:
Thanks for highlighting that caveat 🙌🏻 
I didn't consider that `ua-parser-js` is covered by a relatively strict `AGPL-3` licence. I believe that a simple check might be sufficient.

Issue #244, opened 2026-01-06
The selection hijacking by another element in a not-annotatable area isn't covered sufficiently for keyboard users


oleksandr-danylchenko:
## Context
In my app, the selection is hijacked by the Lexical RTE when a user creates a new note. 
The creation is invoked from a selection popup by clicking a button:

<img height="50" alt="Image" src="https://github.com/user-attachments/assets/a274ee1b-5456-4b4d-9c09-09bc08685769" />

After that, the focus goes to the editor's field, _creating a collapsed selection range in the document_. The editing field is located within an element with the `not-annotatable` class, _but within the annotator's `container`._ 

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/e6c5a96e-0105-4996-9557-c0740e8ddeb7" />

Such a placement is dictated by needing access to the `AnnotoriousContext` and its derived hooks.

## Issue
⚠️ When a user presses the button with the `Enter` key, the selection is unexpectedly cleared, and the annotation gets deleted ⚠️

https://github.com/user-attachments/assets/330745d3-102c-4033-82f8-959310bfb668

https://github.com/recogito/text-annotator-js/blob/cae63fdbd333262dad9694adda9288b62620c0cb/packages/text-annotator/src/SelectionHandler.ts#L204-L217
That happens because for keyboard users, the `currentTarget` isn't reset and refers to an actual selection.

Although for the mouse users, when a user clicks on the button, the `currentTarget` is reset here:
https://github.com/recogito/text-annotator-js/blob/cae63fdbd333262dad9694adda9288b62620c0cb/packages/text-annotator/src/SelectionHandler.ts#L360-L364

## Changes Needed
To handle the hijacking within the not-annotatable within the `container`, the following check needs to become stricter:
https://github.com/recogito/text-annotator-js/blob/cae63fdbd333262dad9694adda9288b62620c0cb/packages/text-annotator/src/SelectionHandler.ts#L158-L166

We should check whether all the ranges are located within not-annotatable elements:
```diff
-if (!selectionRanges.some(r => r.intersectsNode(container))) {
+if (selectionRanges.every(r => !isRangeAnnotatable(container, r))) {
```

PR #245, opened 2026-01-06
#244 Made the hijacking check cover cases when editor is within explicitly `not-annotatable` elements

oleksandr-danylchenko:
## Issue
Fixes the https://github.com/recogito/text-annotator-js/issues/244

## Changes Made
Made the hijacking check stricter by dismissing the selection made entirely within the not-annotatable element, even if it's within the `container`.


rsimon:
Sounds good - thanks!

PR #248, opened 2026-01-09
#247 Fixed keyboard selection in Firefox & Safari after user navigates with the `Tab`

### LINKED ISSUE #247: The keyboard selection stops working in Firefox & Safari after user navigates with the `Tab`
## Issue
There's a difference between what browsers pick as the `keyup` event target after pressing `Tab`. Unfortunately, it makes the keyboard selection appear broken in Firefox and Safari ⚠️ 

I went through the following steps:
1. Select some text with a caret
2. Press the `Tab`
3. Move with an arrow from the focused element
4. Make another selection

| Chrome ✅  | Firefox ⚠️  |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/59dee0e1-3dc1-490f-8fcd-65a4e46d1780" /> | <video src="https://github.com/user-attachments/assets/615c463f-c89c-46de-a7af-b17ce0aabbca" /> | 
| The selection after focusing the footnote creates a new annotation | The selection after focusing the footnote isn't registered, and no new annotations are created | 

The trick is that the `keyup` handler registered on the `container` stops working in FF or Safari after the `Tab` navigation:
https://github.com/recogito/text-annotator-js/blob/db57e5b36999d0f5871c0d2325abecc0403dd977/packages/text-annotator/src/SelectionHandler.ts#L519
https://github.com/recogito/text-annotator-js/blob/db57e5b36999d0f5871c0d2325abecc0403dd977/packages/text-annotator/src/SelectionHandler.ts#L420-L431
Instead, the `keyup` event is reported on the `document.body`.

| Chrome ✅  | Firefox ⚠️  |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/4e509260-96a0-47c4-9262-c7551c1659aa" /> | <video src="https://github.com/user-attachments/assets/8cd54aac-88e2-4c18-ba20-caf2e501c762" /> | 
|  |  Image |

oleksandr-danylchenko:
## Issue
Fixes https://github.com/recogito/text-annotator-js/issues/247

## Changes Made
Following an example of the `handleArrowKeyPress`, I added `keyup` event handling to the `document` level. It will properly capture keyboard selection in Firefox/Safari after a user moves their focus with the `Tab`.

## Relations
This PR depends on:
- https://github.com/recogito/text-annotator-js/pull/246

### DIFF
--- a/packages/text-annotator/src/SelectionHandler.ts
+++ b/packages/text-annotator/src/SelectionHandler.ts
@@ -417,12 +417,33 @@ export const createSelectionHandler = (
     selection.userSelect(currentTarget.annotation, clonePointerEvent(evt));
   }
 
+  /**
+   * Handles the completion of keyboard-based text selection when the Shift key is released.
+   *
+   * When a non-collapsed selection exists upon Shift key release, this handler persists
+   * the annotation to the store and programmatically selects it.
+   *
+   * Event handling is restricted to:
+   * - Annotatable elements within the container
+   * - The document body (which receives focus during Tab navigation in Firefox and Safari)
+   *
+   * @see https://github.com/recogito/text-annotator-js/issues/247
+   */
   const onKeyup = (evt: KeyboardEvent) => {
     if (!currentAnnotatingEnabled) return;
 
+    if (
+      evt.repeat ||
+      (
+        evt.target instanceof Node && isNotAnnotatable(container, evt.target) &&
+        evt.target !== document.body
+      )
+    ) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


oleksandr-danylchenko:
Such a change also resolved a related note creation issue. 
When a user selects a text within a not-annotatable RTE using Shift, the `onKeyUp` handler will still be invoked unexpectedly. That may cause the note's dismissal:

https://github.com/user-attachments/assets/2c9dba57-3079-4c25-97b6-dab7b17d45e5

PR #253, opened 2026-04-21
Fixed unexpected overselection on iOS when a user starts selection from the gutter

oleksandr-danylchenko:
## Issue
I noticed that if the `span` highlights contain some entries, starting selection from the gutter may cause Safari to go wild with the handlebars:

https://github.com/user-attachments/assets/465d2c3b-e264-4eb3-8ce5-b3404bcc4010

## Changes Made
Added explicit `user-select: none;` for the `.r6o-span-highlight-layer` element. That made the selection more predictable in iOS.

###### Soomo Staged (22.04.26)


rsimon:
Interesting - so the `pointer-events: none` alone isn't sufficient? Wild... thanks for figuring this out!!

PR #258, opened 2026-04-28
Updated `@annotorious/*` dependencies and aligned repeated packages versions

oleksandr-danylchenko:
## Changes Made
I updated the `@annotorious/core` from the `^3.7.13` to the latest `^3.8.1`. The latter includes a security fix for the `uuid`. And it's good to keep them up to date.


oleksandr-danylchenko:
Noice!

PR #261, opened 2026-05-21
Rendering Performance: Quick Wins

rsimon:
## In this PR

We're seeing performance limitations on large annotated TEI/XML files. (Large meaning, either, "large filesize and complex markup" and / or "many 100s of annotations.)

Making them work smoothly will require a slightly deeper architecture change, to make the core ([text-annotator-state](https://github.com/recogito/text-annotator-js/blob/764b8938b303cf6160855edbb1d4cfed96c174e6/packages/text-annotator/src/state/text-annotator-state.ts) and [spatial-tree](https://github.com/recogito/text-annotator-js/blob/764b8938b303cf6160855edbb1d4cfed96c174e6/packages/text-annotator/src/state/spatial-tree.ts)) support actual lazy rendering.

But before we do that, here are some smaller "quick wins" that are easy to implement, don't impact the current architecture, and already provide a bit of a performance boost.

### Changes

- Replaced uses of NodeIterator with TreeWalker, which is described as faster in various places (see e.g. [here](https://stackoverflow.com/questions/64551229/queryselectorall-vs-nodeiterator-vs-treewalker-fastest-pure-js-flat-dom-iterat)). Probably not a huge impact, but worth doing.
- Main bottleneck addressed: [get-highlight-client-rects](https://github.com/recogito/text-annotator-js/blob/764b8938b303cf6160855edbb1d4cfed96c174e6/packages/text-annotator/src/utils/highlight/get-highlight-client-rects.ts) was wasting lots of loop iterations. The loop now:
  - skips to the start position immediately
  - breaks early when the end position is reached

### DIFF
--- a/packages/text-annotator-tei/src/crosswalk/utils/reanchor.ts
+++ b/packages/text-annotator-tei/src/crosswalk/utils/reanchor.ts
@@ -3,9 +3,9 @@ export const reanchor = (originalNode: Node, parentNode: Node, originalOffset: n
 
   let offset = originalOffset;
 
-  const it = document.createNodeIterator(parentNode, NodeFilter.SHOW_TEXT);
+  const walker = document.createTreeWalker(parentNode, NodeFilter.SHOW_TEXT);
 
-  let currentNode = it.nextNode();
+  let currentNode = walker.nextNode();
 
   let run = true;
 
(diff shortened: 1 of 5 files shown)


oleksandr-danylchenko (commented):
The changes in the `get-highlight-client-rects` required a bit of groking to understand them. But I see how it would be beneficial in cases when a user made a small selection within a relatively large element. Thanks!

rsimon:
That's right - the performance gain, in most practical cases, is minimal. But we have an unfortunate number of cases (all with TEI/XML documents) where selections happen inside a root element with excessively many child nodes, producing many dozens to 100s of wasted loop iterations.

PR #264, opened 2026-05-21
#263 Added selection preservation, even if it protrudes beyond the `container`

### LINKED ISSUE #263: The selection covering text outside of the `container` will always be dismissed
## Context
In the `onSelectionChange` handler, there's a check that makes sure that all the selection ranges are made within the annotatable `container` and on annotatable elements:

https://github.com/recogito/text-annotator-js/blob/9357b2f10f2dc942cf2227668f3b3c114b1e8464/packages/text-annotator/src/selection-handler.ts#L158-L166

https://github.com/recogito/text-annotator-js/blob/9357b2f10f2dc942cf2227668f3b3c114b1e8464/packages/text-annotator/src/utils/dom/is-not-annotatable.ts#L14-L17

https://github.com/recogito/text-annotator-js/blob/9357b2f10f2dc942cf2227668f3b3c114b1e8464/packages/text-annotator/src/utils/dom/is-not-annotatable.ts#L5-L12

## Issue
The `if (!container.contains(node)) return true;` within the `isNotAnnotatable` appears to be a bit _"too aggressive"_.

When a user creates a mixed selection starting/ending on an element outside of the `container`, the generated range _won't pass the `isRangeAnnotatable` check_! That's because the `commonAncestorContainer` will point to an element outside of the `container`.

 It'll cause an unexpected dismissal of such a selection:

## Issue Demo

https://github.com/user-attachments/assets/4a7b4795-9865-4c61-b9b6-fea19670cdaa

oleksandr-danylchenko:
## Issue
Fixes https://github.com/recogito/text-annotator-js/issues/263

## Changes Made
Added range trimming to the start of the selection processing. That ensures the selection made on the `container` will still get preserved. It should pass the `isRangeAnnotatable` check.

That way, the mixed selection won't be unexpectedly dismissed.

## Demo

https://github.com/user-attachments/assets/39338ca3-30b3-482c-8e54-85d2cbecd262

## Relations
This change is related to the following PR, where any content outside the `container` is treated as `not-annotatable`:
- https://github.com/recogito/text-annotator-js/pull/208

###### Soomo Staged 21.05.26

### DIFF
--- a/packages/text-annotator/src/selection-handler.ts
+++ b/packages/text-annotator/src/selection-handler.ts
@@ -155,12 +155,15 @@ export const createSelectionHandler = <T extends TextAnnotation>(
     const selectionRanges =
       Array.from(Array(sel.rangeCount).keys()).map(idx => sel.getRangeAt(idx));
 
+    const containedRanges =
+      selectionRanges.map(r => trimRangeToContainer(r, container));
+
     /**
      * This is to handle cases where the selection is "hijacked" by
      * another element in a not-annotatable area. A rare case in practice.
      * But rich text editors like Quill will do it!
      */
-    if (selectionRanges.every(r => !isRangeAnnotatable(container, r))) {
+    if (containedRanges.every(r => !isRangeAnnotatable(container, r))) {
       currentTarget = undefined;
       return;
     }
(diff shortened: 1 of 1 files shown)


rsimon:
Thanks!!

PR #265, opened 2026-05-21
[JUNK] Added generic type parameter to `HighlightStyleExpression`

oleksandr-danylchenko:
## Issue
`HighlightStyleExpression` was declared as a non-generic type alias where the `I` type parameter was scoped to the function overload inside the union, not to the type alias itself. This made it impossible for consuming apps to parameterize it with a custom annotation type. The obvious workaround:
```ts
type HighlightStyleFn = Extract<HighlightStyleExpression, Function>;
const style: HighlightStyleFn<WebtextsAnnotation> = ...
// TS2315: Type 'HighlightStyleFn' is not generic.
```
fails because `Extract` produces a non-generic function type, the inner `<I>` is erased.

Additionally, even though `TextAnnotatorOptions<I>` and `TextAnnotator<I>` carried a concrete annotation type parameter, style and `setStyle` were still typed as the unparameterized `HighlightStyleExpression`, so the narrow annotation type was never forwarded to styling callbacks.

## Changes Made
`HighlightStyleExpression` is now a proper generic type alias: `HighlightStyleExpression<I extends TextAnnotation = TextAnnotation>`,  with `TextAnnotation` as the default, so existing usages remain unchanged.

The concrete annotation type is now forwarded from `TextAnnotatorOptions<I>` and `TextAnnotator<I>` down to their `style` / `setStyle` members, and through the renderer and painter interfaces. 

`PDFAnnotator` and its React wrapper are similarly updated to use `PDFAnnotation` as the concrete type.

###### Soomo Staged 21.05.26

### DIFF
--- a/packages/text-annotator/src/rendering/base-renderer.ts
+++ b/packages/text-annotator/src/rendering/base-renderer.ts
@@ -17,7 +17,7 @@ import { debounce } from '../utils/events';
 /**
  * The renderer runtime interface.
  */
-export interface Renderer {
+export interface Renderer<I extends TextAnnotation = TextAnnotation> {
 
   destroy(): void;
 
@@ -25,7 +25,7 @@ export interface Renderer {
 
   redraw(force?: boolean): void;
 
-  setStyle(style?: HighlightStyleExpression, id?: string): void;
+  setStyle(style?: HighlightStyleExpression<I>, id?: string): void;
 
   setFilter(filter?: Filter<any>): void;
 
@@ -54,15 +54,15 @@ export type RendererFactory<T extends Annotation> = (
  * implementers can simply implement a painter, and wrap it in the
  * BaseRenderer, which will handle common concerns.
  */
-export interface Painter {
+export interface Painter<I extends TextAnnotation = TextAnnotation> {
 
   destroy(): void;
 
   redraw(
-    highlights:Highlight[], 
-    bounds: ViewportBounds, 
-    style?: HighlightStyleExpression,
-    styleOverrides?: Map<string, HighlightStyleExpression>,
+    highlights: Highlight[],
+    bounds: ViewportBounds,
+    style?: HighlightStyleExpression<I>,
+    styleOverrides?: Map<string, HighlightStyleExpression<I>>,
     force?: boolean
   ): void;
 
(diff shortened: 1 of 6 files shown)


oleksandr-danylchenko:
Hmm... The new `TextAnnotationLike` and the derived types got in the way:
```
src/text-annotator-options.ts:40:36 - error TS2344: Type 'I' does not satisfy the constraint 'TextAnnotation'.
  Type 'TextAnnotationLike' is not assignable to type 'TextAnnotation'.
    The types of 'target.selector' are incompatible between these types.
      Type 'TextSelectorLike[]' is not assignable to type 'TextSelector[]'.
        Type 'TextSelectorLike' is missing the following properties from type 'TextSelector': start, end

40   style?: HighlightStyleExpression<I>;
                                      ~

src/text-annotator.ts:41:22 - error TS2344: Type 'I' does not satisfy the constraint 'TextAnnotation'.
  Type 'TextAnnotationLike' is not assignable to type 'TextAnnotation'.
    The types of 'target.selector' are incompatible between these types.
      Type 'TextSelectorLike[]' is not assignable to type 'TextSelector[]'.
        Type 'TextSelectorLike' is missing the following properties from type 'TextSelector': start, end

41   renderer: Renderer<I>;

...
```

rsimon:
I'm a bit torn about this. I introduced `TextAnnotationLike` this morning, but still need to wrap my head around what should actually be the right type architecture. In fact I'm already considering getting rid of it again. 

- `TEIAnnotation` is definitely not a `TextAnnotation` any more (because it now lacks char offsets)
- Host apps must be allowed to load `TEIAnnotation`s in and they should Just Work.
- Any `TEIAnnotation` generated by Recogito itself is necessarily **also** a `TextAnnotation`, because it always gets created as one and then gets crosswalked to a `TEIAnnotation`. 

Most parts of the core won't ever notice the difference, but some may have to be guarded against expecting `start` and `end`  props on the selector, if that makes sense?

packages/text-annotator/src/rendering/base-renderer.ts
```diff
@@ -39,50 +39,50 @@ export interface RendererEvents {
...
   setVisible(visible: boolean): void;
 
 }
 
-export const createRenderer = <T extends TextAnnotatorState = TextAnnotatorState<TextAnnotation, any>> (
-  painter: Painter,
-  container: HTMLElement, 
+export const createRenderer = <T extends TextAnnotatorState = TextAnnotatorState<TextAnnotation, any>, I extends TextAnnotationLike = TextAnnotation>(
+  painter: Painter<I>,
+  container: HTMLElement,
```
oleksandr-danylchenko, line 75: Ugh... That seems ugly and convoluted. The type of the annotation is actually in the state already

oleksandr-danylchenko:
Yeah, with an alternative "lightweight" annotation, there's quite a lot of churn that gets introduced. Especially if we wanna preserve the type inference through the whole "dependencies" tree of a concrete annotator implementation 💀

oleksandr-danylchenko:
Hey, @rsimon 👋🏻 
I'll close this PR for now and will resort to a temporary `as any` assertion on the consumer's side. Then I can revisit the changes once we have a more certain understanding of the `TextAnnotationLike` role and need[^1]. Otherwise, all that generics gymnastics may only convolute the codebase.

[^1]: Ideally, I'd preserve only the `TextAnnotation` type to prevent the `"Type could be instantiated with a different subtype"` errors I've already faced.

PR #270, opened 2026-05-27
Added keyboard selection preservation, even if it protrudes beyond the `container`

### LINKED ISSUE #267: Uncaught TypeError: can't access property "ctrlKey", asPtr is undefined
## Issue
When I focus on an element outside of the annotatable container after landing the page, the following exception will appear:

It does nothing functionally, but it creates noise in the logs generated on a page.

I suppose that a null guard should be added to the `isAddToCurrentSelect` method, which will check the `evt` presence.
When I simply navigate through the page, it won't be populated `keydown` listener.

## Demo

https://github.com/user-attachments/assets/181187fa-64d7-4079-b8fd-8c8c30424c29

### LINKED ISSUE #269: The selection covering text outside of the `container`, initiated with the keyboard, will always be dismissed
## Relations
This issue is a follow-up for the previously addressed https://github.com/recogito/text-annotator-js/issues/263.

## Issue
When a user **_starts their selection from the text outside_** of the annotatable `container` using the keyboard, the selection range won't produce an annotation.
Although if the selection **_starts from the `container`_**, a new annotation will be created every time.

That's caused by the guards working too aggressively for the mixed ranges: https://github.com/recogito/text-annotator-js/blob/844e45450f77b210e79270aec01d1bc48092d895/packages/text-annotator/src/selection-handler.ts#L436-L447

Image

The `isNotAnnotatable(container, evt.target)` won't let the processing pass.

## Issue Demo

https://github.com/user-attachments/assets/2a355b88-3344-4e9c-959d-5d4fa036b6e3

oleksandr-danylchenko:
## Relations
This PR is based on the https://github.com/recogito/text-annotator-js/pull/268. I'd highly recommend checking it out first 👀 

## Issue
Fixes #269
Fixes https://github.com/recogito/text-annotator-js/issues/267

---

Keyboard-driven selections (Shift+Arrow) that *start* with the caret outside the annotatable container and *extend into* it were silently dropped. The `onKeyup` and `handleArrowKeyPress` handlers guarded against out-of-container events by inspecting the keyboard event's `target` node via `isNotAnnotatable`. Because the caret (and therefore `evt.target`) originated outside the container, the guard rejected the event — even though the resulting selection range legitimately overlapped annotatable content. A special exception for `document.body` (to cover Tab-navigation focus in Firefox/Safari) papered over some cases but didn't address the general problem.

## Changes Made
Replaced the event-target–based `isNotAnnotatable` checks in `onKeyup` and `handleArrowKeyPress` with an intersection-based guard using a new `getIntersectingRanges()` helper. Instead of asking "did this key event fire from inside the container?", the handlers now ask "does the current selection actually overlap the container?" This correctly allows keyboard selections that begin outside but extend into the annotatable area, while still ignoring selections that don't touch it at all.

The inline range-intersection filtering from `onSelectionChange` was also consolidated into the same shared `getIntersectingRanges()` helper for consistency.

### DIFF
--- a/packages/text-annotator/src/selection-handler.ts
+++ b/packages/text-annotator/src/selection-handler.ts
@@ -152,11 +152,8 @@ export const createSelectionHandler = <T extends TextAnnotation>(
      */
     if (!sel?.anchorNode) return;
 
-    const selectionRanges =
-      Array.from(Array(sel.rangeCount).keys()).map(idx => sel.getRangeAt(idx));
-
     // Drop any range that doesn't intersect the annotatable container at all before we attempt to trim it
-    const intersectingRanges = selectionRanges.filter(r => r.intersectsNode(container));
+    const intersectingRanges = getIntersectingRanges();
 
     // The selection should be captured only within the annotatable container
     const containedRanges = intersectingRanges.map(r => trimRangeToContainer(r, container));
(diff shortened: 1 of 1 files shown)


packages/text-annotator/src/selection-handler.ts
```diff
@@ -152,12 +152,13 @@ export const createSelectionHandler = <T extends TextAnnotation>(
      */
     if (!sel?.anchorNode) return;
 
-    const selectionRanges =
-      Array.from(Array(sel.rangeCount).keys()).map(idx => sel.getRangeAt(idx));
+    const intersectingRanges = getIntersectingRanges();
```
rsimon: Something must have gotten mixed up here. Same const defined twice?

oleksandr-danylchenko: Huh, seems like the conflict auto-resolution didn't work correctly. I'll clean it up 👌🏻

PR #271, opened 2026-05-29
Type Model Refactoring

rsimon:
## In this PR

Attempt to clean up the type model after the introduction of `TextAnnotationLike` and TEI annotations that have no `start` and `end` fields in their selectors.

### DIFF
--- a/packages/pdf-annotator/src/utils/annotation/revive-annotation.ts
+++ b/packages/pdf-annotator/src/utils/annotation/revive-annotation.ts
@@ -27,8 +27,7 @@ export const reviveSelector = (selector: PDFSelector | TextSelector): PDFSelecto
       return selector as PDFSelector;
     } else {
       // No pageNumber, but offsetReference element -> crosswalk
-      const { offsetReference } = selector as unknown as RevivedTextSelectorLike;
-      const pageNumber = parseInt(offsetReference!.dataset.pageNumber!);
+      const pageNumber = parseInt(selector.offsetReference!.dataset.pageNumber!);
     
       // @ts-ignore
       return {
--- a/packages/pdf-annotator/src/utils/annotation/revive-annotation.ts
+++ b/packages/pdf-annotator/src/utils/annotation/revive-annotation.ts
@@ -1,4 +1,4 @@
-import type { TextAnnotation, TextAnnotationTarget, TextSelector } from '@recogito/text-annotator';
+import type { RevivedTextSelectorLike, TextAnnotation, TextAnnotationTarget, TextSelector } from '@recogito/text-annotator';
 import type { PDFAnnotation, PDFAnnotationTarget, PDFSelector } from '../../model/core/pdf-annotation';
 
 /**
@@ -27,7 +27,7 @@ export const reviveSelector = (selector: PDFSelector | TextSelector): PDFSelecto
       return selector as PDFSelector;
     } else {
       // No pageNumber, but offsetReference element -> crosswalk
-      const { offsetReference } = selector;
+      const { offsetReference } = selector as unknown as RevivedTextSelectorLike;
       const pageNumber = parseInt(offsetReference!.dataset.pageNumber!);
     
       // @ts-ignore
--- a/packages/text-annotator/src/model/core/text-annotation.ts
+++ b/packages/text-annotator/src/model/core/text-annotation.ts
@@ -1,32 +1,15 @@
 import type { Annotation, AnnotationTarget } from '@annotorious/core';
 
-export interface TextAnnotation extends Annotation {
-
-  target: TextAnnotationTarget;
-
-}
-
-export interface TextAnnotationTarget extends AnnotationTarget {
-
-  selector: TextSelector[];
-
-}
-
-export interface TextSelector {
-
-  id?: string;
-
-  quote: string;
-  
-  start: number;
-
-  end: number;
-
-  range: Range;
-
-  offsetReference?: HTMLElement;
-
-}
+/**
+ * The Text Annotator model is becoming slightly complex. But it seems
+ * justified since we need to dealing with two orthogonal issues:
+ * 
+ * - Annotations come in "revived" and "unrevived" states (with or without 
+ *   resolved DOM elements: Range and offset element)
+ * - Annotations can be "plain" HTML/text annotations with `start` and `end`
+ *   character offset fields, or annotations from extensions (TEI/XML, PDF)
+ *   which handle the serializable anchoring differently (XPath, pages)
+ */
 
 export interface TextAnnotationLike extends Annotation {
 
@@ -46,8 +29,71 @@ export interface TextSelectorLike {
 
   quote: string;
 
+}
+
+export interface RevivedTextSelectorLike extends TextSelectorLike {
+
   range: Range;
 
   offsetReference?: HTMLElement;
 
+}
+
+/**
+ * Core classes - this is what most parts of the core deal with:
+ * a TextAnnotationLike with all-revived selectors
+ */
+export type RevivedTextAnnotationLike<T extends TextAnnotationLike = TextAnnotationLike> = T & {
+
+  target: RevivedTextAnnotationTargetLike;
+
+}
+
+export type RevivedTextAnnotationTargetLike<T extends TextAnnotationTargetLike = TextAnnotationTargetLike> = Omit<T, 'selector'> & {
+
+  selector: RevivedTextSelectorLike[];
+
+}
+
+/**
+ * Default specialization of TextAnnotationLike
+ */
+export interface TextAnnotation extends TextAnnotationLike {
+
+  target: TextAnnotationTarget;
+
+}
+
+export interface TextAnnotationTarget extends TextAnnotationTargetLike {
+
+  selector: TextSelector[];
+
+}
+
+export interface TextSelector extends TextSelectorLike {
+  
+  start: number;
+
+  end: number;
+
+}
+
+export type RevivedTextSelector = TextSelector & RevivedTextSelectorLike;
+
+/**
+ * Utility type guards
+ */
+export const isRevivedAnnotation = (annotation: TextAnnotationLike): annotation is RevivedTextAnnotationLike => 
+  annotation.target.selector.every(isRevived);
+
+export const isRevivedTarget = (target: TextAnnotationTargetLike): target is TextAnnotationTargetLike & {
+  selector: RevivedTextSelectorLike[]
+} => target.selector.every(isRevived);
+
+export function isRevived(selector: TextSelectorLike): selector is RevivedTextSelectorLike;
+export function isRevived(selectors: TextSelectorLike[]): selectors is RevivedTextSelectorLike[];
+export function isRevived(selector: TextSelectorLike | TextSelectorLike[]): boolean {
+  return Array.isArray(selector)
+    ? selector.every(s => 'range' in s && s.range instanceof Range && !s.range.collapsed)
+    : 'range' in selector && selector.range instanceof Range && !selector.range.collapsed;
 }
--- a/packages/text-annotator/src/model/core/text-annotation.ts
+++ b/packages/text-annotator/src/model/core/text-annotation.ts
@@ -45,7 +45,7 @@ export interface RevivedTextSelectorLike extends TextSelectorLike {
  */
 export type RevivedTextAnnotationLike<T extends TextAnnotationLike = TextAnnotationLike> = T & {
 
-  target: RevivedTextAnnotationTargetLike;
+  target: RevivedTextAnnotationTargetLike<T['target']>;
 
 }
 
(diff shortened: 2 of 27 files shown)


packages/text-annotator/src/model/core/text-annotation.ts
```diff
@@ -1,53 +1,99 @@
...
 
-export interface TextAnnotationLike extends Annotation {
+/**
+ * Core classes - this is what most parts of the core deal with:
+ * a TextAnnotationLike with all-revived selectors
+ */
+export type RevivedTextAnnotationLike<T extends TextAnnotationLike = TextAnnotationLike> = T & {
 
-  target: TextAnnotationTargetLike;
+  target: RevivedTextAnnotationTargetLike;
```
oleksandr-danylchenko:
```suggestion
  target: RevivedTextAnnotationTargetLike<T['target']>;
```
Seems like the `T` should be forwarded here too.

packages/text-annotator/src/utils/annotation/scroll-into-view.ts
```diff
@@ -17,9 +17,9 @@ const getScrollParent = (el?: Element | null) => {
 };
 
 // Executes scroll on an annotation with a valid DOM range selector
-const scroll = <I extends TextAnnotation = TextAnnotation>(
+const scroll = <I extends TextAnnotationLike = TextAnnotationLike>(
   store: TextAnnotationStore<I>,
-  target: I['target'],
+  target: RevivedTextAnnotationTargetLike,
```
oleksandr-danylchenko:
```suggestion
  target: RevivedTextAnnotationTargetLike<I['target']>,
```

packages/text-annotator/src/model/core/text-annotation.ts
```diff
@@ -1,53 +1,99 @@
...
+
+/**
+ * Utility type guards
+ */
+export const isRevivedAnnotation = (annotation: TextAnnotationLike): annotation is RevivedTextAnnotationLike => 
+  annotation.target.selector.every(isRevived);
+
+export const isRevivedTarget = (target: TextAnnotationTargetLike): target is TextAnnotationTargetLike & {
+  selector: RevivedTextSelectorLike[]
+} => target.selector.every(isRevived);
```
oleksandr-danylchenko:
```suggestion
export const isRevivedTarget = (target: TextAnnotationTargetLike): target is RevivedTextAnnotationTargetLike => target.selector.every(isRevived);
```
We can make the guard mark the `target` as  `RevivedTextAnnotationTargetLike`. That would allow narrowing down the type with coercions later in the code.

packages/text-annotator/src/utils/annotation/scroll-into-view.ts
```diff
@@ -92,4 +91,5 @@ export const scrollIntoView = <I extends TextAnnotation = TextAnnotation>(
   }
```
oleksandr-danylchenko, line 91: The logic can be re-writte w/o the coercion:
```ts
  if (isRevivedTarget(current.target)) {
    // The first selector is the topmost one as well
    const { range: annoRange } = current.target.selector[0];
    if (annoRange && !annoRange.collapsed) {
      scroll(store, current.target, scrollParent);
      return true;
    }
  } else {
    // Try reviving now to account for lazy rendering
    const revived = reviveTarget(current.target, container);
    const { range: revivedAnnoRange } = revived.selector[0];
    if (revivedAnnoRange && !revivedAnnoRange.collapsed) {
      scroll(store, revived, scrollParent);
      return true;
    }
  }
```
The [previous suggestion](https://github.com/recogito/text-annotator-js/pull/271#discussion_r3324979987) for the `isRevivedTarget` would help to pass the guarded target to the `scroll` directly.

packages/text-annotator/src/selection-handler.ts
```diff
@@ -31,11 +31,11 @@ const SELECTION_KEYS = [
...
 
-export const createSelectionHandler = <T extends TextAnnotation>(
+export const createSelectionHandler = (
   container: HTMLElement,
-  state: TextAnnotatorState<TextAnnotation, any>,
-  lifecycle: Lifecycle<TextAnnotation, any>,
-  options: TextAnnotatorOptions<TextAnnotation, any>
+  state: TextAnnotatorState<RevivedTextAnnotationLike, any>,
+  lifecycle: Lifecycle<TextAnnotationLike, any>,
+  options: TextAnnotatorOptions<TextAnnotationLike, any>
```
oleksandr-danylchenko, line 38:
```suggestion
export const createSelectionHandler = <T extends TextAnnotationLike = TextAnnotationLike>(
  container: HTMLElement,
  state: TextAnnotatorState<RevivedTextAnnotationLike<T>, any>,
  lifecycle: Lifecycle<T, any>,
  options: TextAnnotatorOptions<T, any>
```

Shouldn't the generic type be preserved?

packages/text-annotator/src/state/text-annotator-state.ts
```diff
@@ -30,14 +31,21 @@ export interface TextAnnotatorState<I extends TextAnnotationLike = TextAnnotatio
 
 }
 
-export const createTextAnnotatorState = <I extends TextAnnotation = TextAnnotation, E extends unknown = TextAnnotation>(
+export const createTextAnnotatorState = <I extends TextAnnotationLike = TextAnnotation, E extends unknown = TextAnnotation>(
   container: HTMLElement,
-  opts: TextAnnotatorOptions<I, E>
+  opts: TextAnnotatorOptions<I, E>,
+  selectorReviveFn?: (arg: I['target']['selector'][number], container: HTMLElement) => RevivedTextSelectorLike
```
oleksandr-danylchenko, line 37: Should the `selectorReviveFn` come as part of the `opts`? That seems to be an easier option than tweaking the positional arguments in the future.

packages/text-annotator/src/utils/annotation/revive-annotation.ts
```diff
@@ -13,12 +19,14 @@ export const isRevived = (selector: TextSelectorLike[]) =>
...
-
+export const reviveTextSelector = <T extends TextSelector>(
+  selector: T, 
+  container: HTMLElement
+): RevivedTextSelector  => {
   const { start, end } = selector;
 
-  const offsetReference = selector.offsetReference || container;
-
+  const offsetReference = (selector as unknown as RevivedTextSelector).offsetReference || container;
```
oleksandr-danylchenko: Hmm, that may be misleading. If we accept only the `TextSelector`, that puts a constraint on the caller side. It _won't expect_ that the `offsetReference` will be processed. Therefore, it makes the revival too implicit/magical.

oleksandr-danylchenko: Initially, I wanted to suggest always using the `isRevived` guard before calling the `reviveTextSelector`. However, that would break the runtime because non-revived annotations might also contain the `offsetReference`, but not the `range`.
The `offsetReference` seems like an outlier 🤔

oleksandr-danylchenko: > However, that would break the runtime because non-revived annotations might also contain the offsetReference, but not the range.

Oh wait, we already call it for non-revived only. Then, expecting the `offsetReference` to be present is still a conceptual mistake that might come as misleding from the caller side.

oleksandr-danylchenko (changes requested):
Thanks for such an extensive submission!

The missing differentiation between the revived and non-revived was quite cumbersome and implicit, indeed. I like the fact that now we can express them separately, but at the cost of a bump in mental complexity.

I spotted a few places where the generic for the base annotation model was omitted. I'd like to know whether it can be preserved for clarity and correctness.
Additionally, I noticed that revival logic depends on the `offsetReference` that should only be present in the revived model. That might be misleading for the caller side, which won't expect that the `offsetReference` will even be processed.

packages/text-annotator/src/model/core/text-annotation.ts
rsimon: Yep, good catch, thanks!

packages/text-annotator/src/utils/annotation/scroll-into-view.ts
rsimon: That's the bit that occurred to me only after I had "finished" the type refactoring. It's connected to the issue of the revived annotation having a Range and a resolved offset element - which is a logical flaw as you say. (Revived is really just about the Range. More below.)

packages/text-annotator/src/state/text-annotator-state.ts
rsimon: Yes, definitely. I was planning that as a 2nd step. (It's in fact already implemented that way in the [lazy-reviving](https://github.com/recogito/text-annotator-js/tree/lazy-reviving) branch - which is just a messy playground though.)

packages/text-annotator/src/utils/annotation/revive-annotation.ts
rsimon: Yes, the problem is that Range and (resolved) offset reference are really two different things, and one doesn't necessarily imply the presence of the other. Not yet sure how to best handle this. I wouldn't want to model another "level" inside the model itself. Possibly I might return to the previous model, where the offsetReference is simply an optional field on the `TextAnnotation`.

In any case, this is being used by the PDF extension only, AFAIK, so it can be considered the responsibility of the extension to make sure the offset reference exists and is correctly resolved.

rsimon:
Hi @oleksandr-danylchenko,

thanks for giving it such a thorough review! All good input, and applied most of the change requests.

- Your comment about putting the `selectorReviveFn` into the `opts`: good call, I was planning to put that in a separate PR, along with a small refactoring of the TEI extension, which is the first extension I want to use this with.
- About the `offsetReference`: well-spotted! That's actually a thing I missed in my refactoring, and only realized it was broken when `scrollIntoView` stopped working in the PDF extension. I need to review the code and remind myself what's going on there. But I think this is really just a workaround needed by the PDF extension so far; and it's filled in by the PDF extension, too. It's not really related to the Range reviving process, so probably shouldn't be sitting in the `Revived`-type at all. If you have any thoughts, let me know. But I think I'll move this back as an optional prop in the `TextAnnotationLike` type - more or less like it was before.

oleksandr-danylchenko:
> But I think this is really just a workaround needed by the PDF extension so far

The `offsetReference` is also used in our system that wraps around the `text-annotator-js`. Textual elements are being broken into separate annotatable chunks. So the `offsetReference` is useful for reviving the ranges in a page-agnostic manner.

> so probably shouldn't be sitting in the `Revived`-type

Yeah, that sounds about right to me. The `offsetReference` should be allowed to be present in the non-revived model. 

I realized that our custom plugin piggybacks on this idea in the W3C annotation parser:
```ts
const parseW3CWebtextsAnnotation = (annotation: W3CWebtextsAnnotation, container?: HTMLElement) => {
	const parseResult = parseW3CTextAnnotation<WebtextsAnnotation, W3CWebtextsAnnotation>(annotation);

	const { parsed, error } = parseResult;
	if (!parsed || error) return parseResult;

	for (const selector of parsed.target.selector) {
		const offsetReference = container?.querySelector<HTMLElement>(WEBTEXT_MATCHER(selector.scope));
		if (!offsetReference) {
			return {
				error: Error(`...`)
			};
		}
		selector.offsetReference = offsetReference;
	}

	/**
	 * Revives the annotation's target range before processing.
	 * It should occur after the `offsetReference` has been set.
	 */
	const revived = reviveAnnotation(parsed, container);

...
```

Therefore, placing it in the `TextSelectorLike` should eliminate the conceptual clash.

packages/pdf-annotator/src/utils/annotation/revive-annotation.ts
```diff
@@ -27,7 +27,7 @@ export const reviveSelector = (selector: PDFSelector | TextSelector): PDFSelecto
       return selector as PDFSelector;
     } else {
       // No pageNumber, but offsetReference element -> crosswalk
-      const { offsetReference } = selector;
+      const { offsetReference } = selector as unknown as RevivedTextSelectorLike;
```
oleksandr-danylchenko:
```suggestion
      const { offsetReference } = selector;
```
No longer needed after https://github.com/recogito/text-annotator-js/pull/271/commits/2e4fea15c53481df212365475dce40b6fac38fd0

oleksandr-danylchenko:
```suggestion
      const { offsetReference } = selector;
```
No longer needed after https://github.com/recogito/text-annotator-js/pull/271/commits/2e4fea15c53481df212365475dce40b6fac38fd0

packages/text-annotator/src/state/text-annotator-state.ts
oleksandr-danylchenko: I applies the suggestion in https://github.com/recogito/text-annotator-js/pull/271#discussion_r3325042400

PR #272, opened 2026-06-03
Moved `selectorReviveFn` option from `createTextAnnotatorState` parameter to `TextAnnotatorOptions`

oleksandr-danylchenko:
## Relations
Applies the suggestion from https://github.com/recogito/text-annotator-js/pull/271#discussion_r3325042400

> Should the `selectorReviveFn` come as part of the `opts`? That seems to be an easier option than tweaking the positional arguments in the future.

> Yes, definitely. I was planning that as a 2nd step. (It's in fact already implemented that way in the [lazy-reviving](https://github.com/recogito/text-annotator-js/tree/lazy-reviving) branch - which is just a messy playground though.)

###### Soomo Staged

### DIFF
--- a/packages/text-annotator/src/state/text-annotator-state.ts
+++ b/packages/text-annotator/src/state/text-annotator-state.ts
@@ -12,7 +12,7 @@ import type {
   HoverState, 
 } from '@annotorious/core';
 import { isRevived } from '../model';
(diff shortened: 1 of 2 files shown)


rsimon:
Looks good, thanks!

PR #273, opened 2026-06-03
Added generic type parameter to `HighlightStyleExpression`

oleksandr-danylchenko:
## Issue
`HighlightStyleExpression` was declared as a non-generic type alias where the `I` type parameter was scoped to the function overload inside the union, not to the type alias itself. This made it impossible for consuming apps to parameterize it with a custom annotation type. The obvious workaround:
```ts
type HighlightStyleFn = Extract<HighlightStyleExpression, Function>;
const style: HighlightStyleFn<WebtextsAnnotation> = ...
// TS2315: Type 'HighlightStyleFn' is not generic.
```
fails because `Extract` produces a non-generic function type, the inner `<I>` is erased.

Additionally, even though `TextAnnotatorOptions<I>` and `TextAnnotator<I>` carried a concrete annotation type parameter, style and `setStyle` were still typed as the unparameterized `HighlightStyleExpression`, so the narrow annotation type was never forwarded to styling callbacks.

## Changes Made

- Made `HighlightStyleExpression<I>` a generic type alias (defaults to `TextAnnotationLike`). Previously `I` was scoped to the inner function signature, so the parameter was erased at the alias level and couldn't be specialized or extracted.
- Threaded `I` through `TextAnnotatorOptions.style`, `TextAnnotator.setStyle`, `PDFAnnotator.setStyle`, and the React `PDFAnnotator`'s `style` prop, so the annotator's annotation type now reaches styling callbacks.
- Made the internal `Renderer.setStyle` method-level generic to accept any narrowed expression, with a single cast at the boundary.

Fully backwards-compatible — all new type parameters default to the previously implicit type. No runtime changes.

###### Soomo Staged

### DIFF
--- a/packages/pdf-annotator-react/src/pdf-annotator.tsx
+++ b/packages/pdf-annotator-react/src/pdf-annotator.tsx
@@ -12,7 +12,7 @@ export type PDFAnnotatorProps = TextAnnotatorOptions<PDFAnnotation, PDFAnnotatio
 
   filter?: Filter;
 
-  style?: HighlightStyleExpression
+  style?: HighlightStyleExpression<PDFAnnotation>
 
   pdfUrl: string;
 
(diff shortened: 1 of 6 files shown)


oleksandr-danylchenko:
Thanks for the https://github.com/recogito/text-annotator-js/pull/271, this PR can supersede the https://github.com/recogito/text-annotator-js/pull/265.

rsimon:
Thanks, looks great!

PR #277, opened 2026-06-11
Support selector splitting in `selectorReviveFn`

oleksandr-danylchenko:
## Issue

`selectorReviveFn` currently has to revive each stored selector into exactly one DOM range. However, in some consuming applications, there's a need to split that single selector into pieces based on the content.
The motivation and the original discussion are in https://github.com/annotorious/annotorious/pull/607#issuecomment-4680742386.

## Changes Made

- Widened the `selectorReviveFn` option signature to return `RevivedTextSelectorLike | RevivedTextSelectorLike[]`, so one stored selector can fan out into multiple revived selectors when needed.
- Updated `reviveTarget` / `reviveAnnotation` to splat the returned arrays via `.flatMap`, producing a flat list of revived selectors on the target.
- Refactored `spatial-tree.ts` so revival is no longer its concern:
  - `toItems` now reads `s.range` directly from already-revived selectors instead of re-running the revive function on every `insert` / `update` / `set` / `recalculate`. Beyond saving work, this sidesteps the old `(...)?.range` access that would have silently dropped array returns from a splitting `selectorReviveFn`.
  - `createSpatialTree` no longer takes a `selectorReviveFn` parameter — the tree is decoupled from the option's signature.
  - `insert`, `update`, and `set` are typed `RevivedTextAnnotationTargetLike`. `remove` stays on the permissive `TextAnnotationTargetLike` since it only uses `target.annotation` for index lookup and never reaches `toItems`.
- Updated the spatial tree's caller in `text-annotator-state.ts` accordingly. Revival continues to happen exactly once per write — upstream in `addAnnotation` / `syncAnnotations` / `bulkUpsertAnnotations` / `updateTarget` — before targets reach the tree.

###### Soomo Staged

### DIFF
--- a/packages/text-annotator/src/state/text-annotator-state.ts
+++ b/packages/text-annotator/src/state/text-annotator-state.ts
@@ -12,7 +12,13 @@ import type {
   HoverState, 
 } from '@annotorious/core';
 import { isRevived } from '../model';
-import type { RevivedTextAnnotationLike, TextAnnotation, TextAnnotationLike, TextAnnotationTargetLike } from '../model';
+import type {
+  RevivedTextAnnotationLike,
+  RevivedTextAnnotationTargetLike,
+  TextAnnotation,
+  TextAnnotationLike,
+  TextAnnotationTargetLike
+} from '../model';
 import { createSpatialTree, type SpatialTreeEvents } from '../state/spatial-tree';
 import type { AnnotationRects, TextAnnotationStore } from '../state/text-annotation-store';
 import { reviveAnnotation, reviveTarget } from '../utils/annotation';
@@ -42,8 +48,7 @@ export const createTextAnnotatorState = <I extends TextAnnotationLike = TextAnno
     store,
     container,
     opts.mergeHighlights?.horizontalTolerance,
-    opts.mergeHighlights?.verticalTolerance,
-    opts.selectorReviveFn
+    opts.mergeHighlights?.verticalTolerance
   );
 
   const selection = createSelectionState<I, E>(store, opts.userSelectAction, opts.adapter);
@@ -151,10 +156,10 @@ export const createTextAnnotatorState = <I extends TextAnnotationLike = TextAnno
       deleted.forEach(a => tree.remove(a.target));
 
     if (created.length > 0)
-      tree.set(created.map(a => a.target), false);
+      tree.set(created.map(a => a.target as RevivedTextAnnotationTargetLike), false);
 
     if (updated?.length > 0)
-      updated.forEach(({ newValue }) => tree.update(newValue.target));
+      updated.forEach(({ newValue }) => tree.update(newValue.target as RevivedTextAnnotationTargetLike));
   });
 
   return {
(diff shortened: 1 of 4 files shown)


oleksandr-danylchenko:
I asked Claude to trace the usage of the `toItems` method in the spatial tree. Fortunately, as I suspected, calling the revival function there may be excessive:

> Tracing the call graph from `toItems` outward:
> 
> **Inside `spatial-tree.ts`**, `toItems` is only reached via `insert` (line 99), `set` (line 125), `update` → `insert` (line 116), and `recalculate` → `set` (line 213).
> 
> **External callers** of the tree (in `text-annotator-state.ts`):
> - `tree.set(created.map(a => a.target), false)` (line 154) — `created` is filtered by `isRevived(a.target.selector)` at line 147
> - `tree.update(newValue.target)` (line 157) — `updated` is filtered by `isRevived(...)` at line 148
> - `tree.recalculate()` (line 142) — calls `set(store.all()...)` with no inline filter, but every write path into the store revives first:
>   - `addAnnotation` calls `reviveAnnotation` (line 57) before `store.addAnnotation`
>   - `syncAnnotations` / `bulkUpsertAnnotations` / `bulkAddAnnotations` go through `reviveAll` (lines 67, 77, 87)
>   - `updateTarget` / `bulkUpdateTargets` call `reviveTarget` (lines 103, 108)
> 
> So in practice, **no** — every target that reaches `toItems` has already been revived once.
> 
> Caveats worth knowing for the work on this branch:
> 
> 1. **The type signature doesn't enforce it.** `toItems(target: TextAnnotationTargetLike, ...)` accepts unrevived targets, and the store returned at `text-annotator-state.ts:161` spreads `...store` — so callers can still hit base-store methods that bypass the revive wrappers. They'd then be filtered out only at the observer (lines 147-148), but `recalculate()` reads `store.all()` without an `isRevived` filter, so an unrevived entry that snuck into the store would reach `toItems`.
> 
> 2. **`toItems` re-runs the revive function on every call**, even though the input is already revived. `reviveTextSelector` is idempotent (early-return at `revive-annotation.ts:26`), but a user-supplied `selectorReviveFn` has no such guarantee — and on this branch where it can now return an array (splitting), you should decide whether the contract requires it to handle already-revived input.
> 
> 3. **Unrelated but worth flagging**: line 60-63 — `(selectorReviveFn ? ... : ...)?.range` — won't behave correctly once `selectorReviveFn` is allowed to return `RevivedTextSelectorLike[]`. `.range` on an array is `undefined`, so the array return path silently drops all rects. That's the open work the branch name implies.

packages/text-annotator/src/state/text-annotator-state.ts
```diff
@@ -151,10 +156,10 @@ export const createTextAnnotatorState = <I extends TextAnnotationLike = TextAnno
       deleted.forEach(a => tree.remove(a.target));
 
     if (created.length > 0)
-      tree.set(created.map(a => a.target), false);
+      tree.set(created.map(a => a.target as RevivedTextAnnotationTargetLike), false);
 
     if (updated?.length > 0)
-      updated.forEach(({ newValue }) => tree.update(newValue.target));
+      updated.forEach(({ newValue }) => tree.update(newValue.target as RevivedTextAnnotationTargetLike));
```
oleksandr-danylchenko, line 162: Using coercion is safe here because the `created` and `updated` arrays are filtered with the `isRevived` check first.

rsimon:
I'll have to look into this tomorrow. But, yes, I agree that the store should own the reviving, and the tree shouldn't be concerned. 👍

PR #278, opened 2026-06-12
Added revival step for `updateAnnotation` wrapper before persisting annotation to store

oleksandr-danylchenko:
## Issue
The `updateAnnotation` method misses the revival step before delegating the persistence to the base A9S method.

Due to the "revival" being performed both for attaining the `range` and performing the crosswalk, we'll miss the latter on the update. However, consumers might expect it to happen on change of the selected content.

Previously, that issue wasn't noticeable due to using a monkey-patched `updateAnnotation` through a plugin. But the issue appeared after I switched to the newer `selectorReviveFn` approach.

---

While working on the PR, I noticed that the `updateTarget` and `bulkUpdateTargets` didn't forward the `opts.selectorReviveFn`. That was breaking the selection logic when a custom `selectorReviveFn` is passed.

###### Soomo Staged

### DIFF
--- a/packages/text-annotator/src/state/text-annotator-state.ts
+++ b/packages/text-annotator/src/state/text-annotator-state.ts
@@ -115,12 +115,12 @@ export const createTextAnnotatorState = <I extends TextAnnotationLike = TextAnno
   }
 
   const updateTarget = (target: TextAnnotationTargetLike, origin = Origin.LOCAL) => {
-    const revived = reviveTarget(target, container);
+    const revived = reviveTarget(target, container, opts.selectorReviveFn);
     store.updateTarget(revived, origin);
   }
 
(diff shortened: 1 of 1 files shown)


rsimon:
Yikes - many thanks!

oleksandr-danylchenko:
The great news is that the `selectorReviveFn` is robust enough now! I migrated from the plugin approach, and today those changes passed the QC on our side 🎉 

---

Additionally, now the selection is correctly preserved when new remote annotations arrive, due to the `syncAnnotations` incorporation.

rsimon:
Woohoo 🎉! Thanks so much this. I'm happy that everything keeps working on your end, and I'm truly grateful for your work in troubleshooting/hardening this! (And, well, and patching all the holes that my initial version had ;-) 

The code isn't actually in production yet on our end, so knowing that it's already passed through your QC is super-valuable for us!

PR #279, opened 2026-06-17
Add `hideAnnotations` prop to React text/TEI annotators

blms:
### In this PR

Per https://github.com/recogito/recogito-studio-sdk/pull/13#pullrequestreview-4513063408:
- Add `hideAnnotations` prop to React wrappers for text/TEI annotators, using `setVisible` on the annotator instance

### Questions

Is it ok that this will only work for now on the span renderer? It seems `setVisible` is defined there but not on the CSS highlight renderer.

### DIFF
--- a/packages/text-annotator-react/src/tei/tei-annotator.tsx
+++ b/packages/text-annotator-react/src/tei/tei-annotator.tsx
@@ -13,6 +13,8 @@ export type TEIAnnotatorProps = TextAnnotatorOptions<TEIAnnotation, unknown> & {
 
   filter?: Filter;
 
+  hideAnnotations?: boolean;
+
 }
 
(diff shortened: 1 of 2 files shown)


oleksandr-danylchenko (approved):

rsimon:
Looks good, thanks!

> Is it ok that this will only work for now on the span renderer? It seems setVisible is defined there but not on the CSS highlight renderer.

IMO that's ok. The interface is correctly defined, and the CSS renderer technically implements the function. Obviously, it just logs a message ATM. But no-one uses the CSS renderer as far as aware, and I'd consider it "alpha" at best. I'll merge and cut a new release!

rsimon (approved):

PR #284, opened 2026-08-17
Remove the scroll listener with the capture flag it was registered with

### LINKED ISSUE #283: destroy() leaks the document scroll listener: removeEventListener omits the capture flag
### Summary

`destroy()` does not remove the renderer's `scroll` listener, because the `removeEventListener` call omits the `capture` flag it was registered with. Every destroyed annotator therefore leaves a live, document-level listener behind, and each orphan keeps calling `redraw(true)` — a *forced* full highlight-layer rebuild — on every scroll event in the document.

`packages/text-annotator/src/rendering/base-renderer.ts`:

```ts
// :191 — registered in the capture phase
document.addEventListener('scroll', onScroll, { capture: true, passive: true });

// :232 — removed without it, so the DOM never matches this registration
document.removeEventListener('scroll', onScroll);
```

`removeEventListener` only matches a registration whose capture flag is equal, so this call is a no-op.

Present in 4.2.5 (current release, bundled at `dist/text-annotator.es.js:331` / `:343`) and on `main` @ 092be72.

Interestingly, `packages/text-annotator-react/src/text-annotation-popup/text-annotation-popup.tsx:160-164` gets this right — it passes `true` to both calls.

### Why it hurts

Because it accumulates. Anything that mounts and unmounts annotators over a session — collapsing/expanding a panel, re-seeding an instance, client-side route changes, or several annotators on one page — piles up orphans that are invisible to the app but permanently attached to the document. Each one, per scroll event, still runs the full `redraw(true)` path: `getBoundingClientRect()` on the (now detached) container, an rbush intersection query, and a `rAF` that clears and rebuilds the highlight layer.

I found this while profiling scroll jank in a Chrome side panel that mounts one annotator per output card. The library's scroll redraw was already the largest single cost in the trace at *one* live instance (its 10 ms trailing debounce expires on every 16.7 ms frame, so it is effectively per-frame); with orphans accumulating, it multiplies.

### Repro

`node repro.mjs` — jsdom, 5 mount/destroy cycles, counting what stays on the document. Removal is instrumented to mirror the DOM's own matching rule (same listener **and** same capture flag), so a mismatched `removeEventListener` behaves exactly as it does in a browser: silently doing nothing.

<details>
<summary>repro.mjs</summary>

```js
import { JSDOM } from 'jsdom';

const dom = new JSDOM('<!doctype html><html><body><div id="a">Some annotatable text here.</div></body></html>', {
  pretendToBeVisual: true,
});
const { window } = dom;

// --- minimal browser globals the library expects -----------------------------
globalThis.window = window;
globalThis.document = window.document;
globalThis.DOMRect = window.DOMRect;
globalThis.Node = window.Node;
globalThis.Range = window.Range;
globalThis.Event = window.Event;
globalThis.MutationObserver = window.MutationObserver;
globalThis.getSelection = () => window.getSelection();
globalThis.requestAnimationFrame = (cb) => setTimeout(() => cb(Date.now()), 0);
globalThis.cancelAnimationFrame = (id) => clearTimeout(id);
Object.defineProperty(globalThis, 'navigator', { value: window.navigator, configurable: true });
globalThis.ResizeObserver = window.ResizeObserver = class {
  observe() {}
  unobserve() {}
  disconnect() {}
};
// jsdom has no layout engine; the library only needs these to not throw.
window.Range.prototype.getClientRects = () => [];
window.Range.prototype.getBoundingClientRect = () => new window.DOMRect();

// --- instrument document scroll registrations -------------------------------
const origAdd = document.addEventListener.bind(document);
const origRemove = document.removeEventListener.bind(document);
const captureOf = (options) => (typeof options === 'boolean' ? options : !!options?.capture);
/** listener -> { capture, wrapper, calls } for every live scroll registration. */
const live = new Map();

document.addEventListener = (type, listener, options) => {
  if (type !== 'scroll' || typeof listener !== 'function') return origAdd(type, listener, options);
  const entry = { capture: captureOf(options), calls: 0 };
  entry.wrapper = (event) => {
    entry.calls++;
    return listener(event);
  };
  live.set(listener, entry);
  return origAdd(type, entry.wrapper, options);
};

document.removeEventListener = (type, listener, options) => {
  if (type !== 'scroll' || typeof listener !== 'function') return origRemove(type, listener, options);
  const entry = live.get(listener);
  if (!entry || entry.capture !== captureOf(options)) return; // no match — the DOM keeps the listener
  live.delete(listener);
  return origRemove(type, entry.wrapper, options);
};

// --- mount + destroy in a loop ----------------------------------------------
const { createTextAnnotator } = await import('@recogito/text-annotator');

const container = document.getElementById('a');
const ROUNDS = 5;
for (let i = 0; i < ROUNDS; i++) {
  createTextAnnotator(container, { renderer: 'SPANS' }).destroy();
}

console.log(`mount + destroy cycles:                  ${ROUNDS}`);
console.log(`scroll listeners still on the document:  ${live.size}`);

// One scroll event, to show the orphans are not merely registered but running.
container.dispatchEvent(new window.Event('scroll'));
console.log(`orphans invoked by a single scroll event: ${[...live.values()].filter((e) => e.calls > 0).length}`);

// The capture flag is the only thing missing — removing with it clears them all.
const leaked = live.size;
for (const listener of [...live.keys()]) document.removeEventListener('scroll', listener, true);
console.log(`after re-removing with { capture: true }: ${live.size} (was ${leaked})`);
```

</details>

Output with 4.2.5:

```
mount + destroy cycles:                  5
scroll listeners still on the document:  5
orphans invoked by a single scroll event: 5

after re-removing with { capture: true }: 0 (was 5)
```

The last line is the whole diagnosis: nothing about the listener identity is wrong, only the flag.

### Suggested fix

```diff
-  document.removeEventListener('scroll', onScroll);
+  document.removeEventListener('scroll', onScroll, { capture: true });
```

Happy to send this as a PR if useful.

### Side note in the same `destroy()`

`onResize` is a `debounce(...)` and `onResize.clear()` is commented out at `base-renderer.ts:234`. A resize debounced just before `destroy()` can therefore still fire afterwards and run `store.recalculatePositions()` + `redraw()` against a destroyed painter. Much smaller than the leak above (bounded, one-shot), but it looks unintentional given the commented-out line.

### Environment

- `@recogito/text-annotator` 4.2.5 (latest on npm), `SPANS` renderer
- Found by reading the source while profiling scroll performance in a Chromium side panel; reproduced under jsdom as above. Not renderer-specific — the listener lives in `base-renderer`.

reorx:
Fixes #283.

### The bug

`base-renderer.ts` registers its scroll listener in the capture phase but removes it without the flag:

```ts
// :191
document.addEventListener('scroll', onScroll, { capture: true, passive: true });

// :232 — before this PR
document.removeEventListener('scroll', onScroll);
```

`removeEventListener` only matches a registration whose capture flag is equal, so that call was a no-op: **every `destroy()`ed renderer left its scroll listener attached to the document.** Each orphan keeps calling `redraw(true)` — a forced, full highlight-layer rebuild — on every scroll event in the page, so the cost accumulates over a session with every mount/destroy cycle (collapsing a panel, re-creating an annotator, client-side route changes, several annotators on one page).

### The change

One line, plus a note so it does not regress:

```diff
-    document.removeEventListener('scroll', onScroll);
+    // Note: the capture flag must match the one used on registration,
+    // otherwise this call matches no listener and silently does nothing.
+    document.removeEventListener('scroll', onScroll, { capture: true });
```

### The test

`test/rendering/base-renderer.test.ts` drives `createRenderer` with a stub painter/state and asserts the observable behaviour rather than the listener bookkeeping:

- **control** — a scroll event *does* reach the painter while the renderer is mounted, so the harness is known to be live;
- **regression** — after `destroy()`, a scroll event must not reach the painter.

Verified in both directions: the regression test fails on `main` (`painter.redraw` is called once after `destroy()`) and passes with the fix. The control test passes either way.

`npx vitest run` in `packages/text-annotator` → 5 passed. `tsc -p tsconfig.build.json` clean. `package-lock.json` deliberately untouched.

### Not included

`onResize.clear()` is commented out two lines below (`base-renderer.ts:234`), so a resize debounced just before `destroy()` can still fire afterwards and run `store.recalculatePositions()` + `redraw()` against a destroyed painter. That one needs a `clear()` on the local `debounce` helper (the current implementation does not expose one), so it felt out of scope here — happy to follow up if you want it fixed too.

### DIFF
--- a/packages/text-annotator/src/rendering/base-renderer.ts
+++ b/packages/text-annotator/src/rendering/base-renderer.ts
@@ -229,7 +229,9 @@ export const createRenderer = <T extends TextAnnotatorState<TextAnnotationLike, 
     unsubscribeSelection();
     unsubscribeHover();
 
-    document.removeEventListener('scroll', onScroll);
+    // Note: the capture flag must match the one used on registration,
+    // otherwise this call matches no listener and silently does nothing.
+    document.removeEventListener('scroll', onScroll, { capture: true });
 
     // onResize.clear();
     window.removeEventListener('resize', onResize);
--- a/packages/text-annotator/test/rendering/base-renderer.test.ts
+++ b/packages/text-annotator/test/rendering/base-renderer.test.ts
@@ -1,0 +1,87 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { createRenderer, type Painter } from '../../src/rendering/base-renderer';
+
+/**
+ * The renderer redraws through a 10ms debounce and a rAF, so give both a
+ * chance to run before asserting.
+ */
+const flushRedraw = () => new Promise(resolve => setTimeout(resolve, 50));
+
+const createPainter = () => ({
+  destroy: vi.fn(),
+  redraw: vi.fn(),
+  setVisible: vi.fn()
+}) satisfies Painter;
+
+const createState = () => ({
+  store: {
+    observe: vi.fn(),
+    unobserve: vi.fn(),
+    getAt: vi.fn(),
+    getIntersecting: vi.fn(() => []),
+    recalculatePositions: vi.fn()
+  },
+  selection: {
+    subscribe: vi.fn(() => vi.fn()),
+    selected: [],
+    evalSelectAction: vi.fn()
+  },
+  hover: {
+    subscribe: vi.fn(() => vi.fn()),
+    current: undefined,
+    set: vi.fn()
+  }
+});
+
+const viewport = { set: vi.fn() };
+
+describe('createRenderer', () => {
+
+  let container: HTMLElement;
+
+  beforeEach(() => {
+    // jsdom has no ResizeObserver.
+    global.ResizeObserver = class {
+      observe() {}
+      unobserve() {}
+      disconnect() {}
+    };
+
+    container = document.createElement('div');
+    document.body.appendChild(container);
+  });
+
+  afterEach(() => {
+    container.remove();
+  });
+
+  it('should redraw when the page scrolls', async () => {
+    const painter = createPainter();
+    const renderer = createRenderer(painter, container, createState() as any, viewport as any);
+
+    container.dispatchEvent(new Event('scroll'));
+    await flushRedraw();
+
+    expect(painter.redraw).toHaveBeenCalled();
+
+    renderer.destroy();
+  });
+
+  it('should stop redrawing on scroll after destroy', async () => {
+    // Regression test: the scroll listener is registered with { capture: true },
+    // so destroy() has to remove it with the same flag. Without it the listener
+    // stays on the document forever, and every destroyed renderer keeps forcing
+    // a full redraw on every scroll event in the page.
+    const painter = createPainter();
+    const renderer = createRenderer(painter, container, createState() as any, viewport as any);
+
+    renderer.destroy();
+    painter.redraw.mockClear();
+
+    container.dispatchEvent(new Event('scroll'));
+    await flushRedraw();
+
+    expect(painter.redraw).not.toHaveBeenCalled();
+  });
+
+});


oleksandr-danylchenko:
Awesome finding! Thanks for the contribution!

oleksandr-danylchenko (approved):

rsimon (approved):
Thanks for catching, looks great!

Issue #288, opened 2026-09-09
PDF with embedded JBIG2 failing, probably since PDF.js v5.4.624


blms:
Hey @rsimon! A client of Recogito uploaded a PDF with embedded JBIG2 images, and the pdf annotator seems to be failing to render it:  [Quinn_compressed.pdf](https://github.com/user-attachments/files/32027578/Quinn_compressed.pdf) (testing under recogito client with v4.3.0 of this package)

It looks like this may have been introduced by PDF.js switching to a WASM decoder for JBIG2 in https://github.com/mozilla/pdf.js/releases/tag/v5.4.624.

rsimon:
Hi @blms, I'm investigating... the PDF still works when viewed in Firefox (which supposedly uses PDF.js underneath the hood); and when I open it in the PDF Annotator, I'm getting an error complaining about a missing `wasmUrl` parameter. So, must be definitely related to the change you mention, and hopefully just a small thing to fix. I'll report back!

rsimon:
Hi @blms, done! The change was indeed minimal, and the actual fix is to make sure the WASM file gets copied to the right locations. I made those modifications in this repo (copying the WASM files into the dev root for local testing and into `dist` for redistribution); and also have a draft PR up in `recogito-client`, which modifies the PDF.js asset copy script accordingly, and which will work once I cut a new release of the `pdf-annotator`.

rsimon:
New release is out: [v4.3.1](https://www.npmjs.com/package/@recogito/pdf-annotator/v/4.3.1)

PR #289, opened 2026-09-10
PDF with embedded JBIG2

rsimon:
## In this PR

This PR is in response to issue #288: 
- The PDF annotator now supports an additional `wasmUrl` option, to configure the path where the PDFium JBig2 decoder WASM is hosted, which is [now required to support PDFs with JBIG2 images](https://github.com/mozilla/pdf.js/releases/tag/v5.4.624). Per default, `wasmUrl` points to the app root path `/`.
- The assets copy script was modified to copy the runtime files (WASM + JS fallback) into the package root (for the interactive test environment); and to copy runtime & license files into the `dist/` folder for redistribution.


blms (approved):
Looks good, thanks for tackling this so quickly!

